PR #73, opened 2015-07-13
First Pull test ;)

AlgorithmsOfMathdestruction:
Added several features to spellbot:

-explicit equip/discard task for use when using "use" task with toggle is unsure
-eat task having "feeding" keyword for name attr to eat anything eatable from invent 
-added loop and endloop task to repeat a bunch of task commands. Implemented with a stack of stacks to be able to place loops inside loops
-added multicast task to repeat cast + sleep without loop command
-added "cap" attr to cast to trigger "onmax" before 99%, no value or zero runs to 99%, enter cap 100 to repeat already maxed out spell like "create food"
-added buy commands to buy items from npc (incldung vault). added attr "invmax" to specify a maximum of the selected item in inventory (prevent overbuy on out of mana etc)
-added "recover" task with attr "stat" taking hp, mana or vigor. rests char until stat is full. (maybe sometimes failes to stop resting .. don't know why yet)

-added good bunch of commented example templates

### DIFF
--- a/Meridian59.Bot.Spell/SpellBotClient.cs
+++ b/Meridian59.Bot.Spell/SpellBotClient.cs
@@ -801,7 +801,7 @@ namespace Meridian59.Bot.Spell
 
             // Cap to maxamount
             InventoryObject existingStock = Data.InventoryObjects.GetItemByName(targetItem.Name, false);
-            if (existingStock.Name.Equals(targetItem.Name))
+            if (existingStock != null && existingStock.Name.Equals(targetItem.Name))
             {
                 if (existingStock.Count > Task.MaxAmount)
                 {


Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -57,6 +67,15 @@ public override void Init()
...
             tickSleepUntil = GameTick.Current + STARTUPSLEEP;
+
+            // setup list of known food and nutrition
+            foodList = new Dictionary<string, byte>() { 
+            {"bunch of grapes",7}, {"loaf of bread",20}, {"apple",10}, {"wheel of cheese",30},
+            {"mug of stout",6}, {"water skin",3}, {"goblet of wine",6}, {"goblet of Ale",3},
+            {"mugs of bew",3}, {"inky-cap mushroom",50},{"edible mushroom",5}, {"meat pie",30},
+            {"bowl of soup",9}, {"bowl of stew",15}, {"fortune cookie",1}, {"slice of pork",9},
+            {"turkey leg",15}, {"drum stick",9}, {"drumstick",9}, {"spider eyes",9},
+            {"spider eye",9},{"copper pekonch mugs",3} };
```
cyberjunk: If possible please define these food strings in the core-library (File: ResourceStrings.cs):
static class Meridian59.Common.Constants.Items.Food - there are already a few entries in it which can come handy in a bot for comparison.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -159,12 +178,75 @@ protected override void ProcessCommand(uint PartnerID, string[] Words)
...
+        protected bool ProcessIsStatRecovered(string stat)
+        {
+            int current = 0, max = 0;
+
+            switch(stat)
+            {
+                case "HP":
+                    {
+                        current = Data.AvatarCondition.GetItemByNum(1).ValueCurrent;
+                        max = Data.AvatarCondition.GetItemByNum(1).ValueMaximum;
```
cyberjunk: Please use the statnum constants from Constants.cs (Common.Constants) rather than the direct values in GetItemByNum(1).There is StatNums.HITPOINTS (=1), StatNums.MANA (=2),...

Also GetItemByNum() may return NULL in case this condition information is not available (may occur during serversaves for a very short interval I think). So you better check the return of GetItemByNum() against NULL before you access it (by .ValueCurrent and .ValueMaximum)

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -371,6 +539,115 @@ protected void DoCast(BotTaskCast Task)
...
+                    SendReqUseMessage(bestEat.ID);
+                    Log("BOT", "Executed task 'feeding' " + bestEat.Name);
+                    DoSleep(new BotTaskSleep(1000));
+                    return;
+                }
+                
+            }
+            else // eat concrete food item
+            {
+                if (Data.ChatMessages.LastAddedItem.FullString.ToLower().Equals("You are too full to eat."))
```
cyberjunk: 'You are too full to eat.' is already defined in the core library:
Please have a look at (ChatSubStrings.cs) - Meridian59.Common.Constants.ChatSubStrings

If you have interesting chatstrings/substrings to interpret chat receveid from the server, add it there please :-)

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -261,6 +425,10 @@ protected void DoCast(BotTaskCast Task)
             SpellObject spellObject = null;
             StatList spellStat = null;
 
+            // test for selfcast
+            if (Task.Target.ToLower().Equals("self"))
+                Task.Target = Data.AvatarObject.Name;
```
cyberjunk: There are some cases when Data.AvatarObject value can be NULL.
It is set when the client receives the RoomObjects list, that happens slighty after the room is loaded and quite after the client is started up / logged in. I recommend to check Data.AvatarObject for NULL before accessing it directly (by .Name)

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -371,6 +539,115 @@ protected void DoCast(BotTaskCast Task)
...
+                if (foodObject == null)
+                {
+                    // log
+                    Log("WARN", "Cant execute task 'eat'. Item " + Task.Name + " not found.");
+                    return;
+                }
+
+                // if food is known internally test if useage would be waste
+                if(foodList.ContainsKey(foodObject.Name))
+                    if((200 - Data.VigorPoints) < foodList[foodObject.Name] )
```
cyberjunk: Vigor maximum of 200 is defined in Constants.cs (Common.Constants) -> StatNumsValues.MAXVIGOR

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -159,12 +178,75 @@ protected override void ProcessCommand(uint PartnerID, string[] Words)
...
+                        break;
+                    }
+                case "VIG":
+                    {
+                        current = Data.AvatarCondition.GetItemByNum(3).ValueCurrent;
+                        
+                        SkillList tl = Data.AvatarSkills;
+                        StatList[] sl = tl.ToArray();
+
+                        max = 80; // default
```
cyberjunk: Please add the '80' in Common.Constants.StatNumsValues... like DEFAULTVIGOR.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -172,41 +254,93 @@ public override void Update()
...
-                // get next task
-                currentTask = Config.GetNextTask();
-
-                // no task?
-                if (currentTask == null)
+                //// Emergency log
+                //if (Data.HitPoints < (Data.AvatarCondition.GetItemByNum(1).ValueMaximum / 3))
+                //{
+                //    DoCast(new BotTaskCast("phase", "", "", "", 100));
+                //}
```
cyberjunk: Is that supposed to be some kind of deactivated auto-phase feature? :-)
You probably want to remove that from a pullrequest or implement it correctly.

Also if you do stuff which is not compatible with original 101 stuff (like phase) - please try using "#if !VANILLA" statements to switch code for each variant of the game accordingly.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -402,6 +678,190 @@ protected void DoUse(BotTaskUse Task)
...
+        /// </summary>
+        /// <param name="Task"></param>
+        protected void DoEquip(BotTaskEquip Task)
+        {
+            // try to get the item from the inventory
+                ObjectBaseList<InventoryObject> foundItems = 
+                    Data.InventoryObjects.GetItemsByName(Task.Name, false);
+
+            // item not found
+            if (foundItems == null)
```
cyberjunk: GetItemsByName() above will never return null - however it may return a list with 0 entries.
So the code within the if() will never be reached - check for (foundItems.Count == 0) instead of null

Meridian59.Bot.Spell/configuration.xml
```diff
@@ -129,11 +60,13 @@
   
   <!-- 
     CONNECTION:
+      character: name of your character to log in
       majorverson: see client menu help/about
       minorversion: see client menu help/about
       logfile: the bot will log also to this file (cleaned at startup, "" to disable)
   -->
   <connection
+    character="CHAR_NAME"
```
cyberjunk: It seems you have not migrated your configuration.xml to the latest changes in my GIT.
The character name to login is not defined this way anymore - instead it is defined together with the connection-entry (username/password), see above.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -228,6 +362,36 @@ protected void DoRest(BotTaskRest Task)
...
+        protected void DoRecover(BotTaskRecover Task)
+        {
+            switch (Task.Stat.ToLower())
+            {
+                case "hp":
+                case "hps":
+                case "hitpoint":
+                case "lp":
+                case "lebenspunkte":
+                case "hitpoints": if (Data.HitPoints >= Data.AvatarCondition.GetItemByNum(1).ValueMaximum) return; recoveringHp = true; break;
```
cyberjunk: Same as mentioned somewhere else, please use the constants from Constants.cs StatNums in the GetItemByNum() calls and handle a possible null return.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -402,6 +678,190 @@ protected void DoUse(BotTaskUse Task)
...
+
+            TradeOfferObject targetItem = null;
+
+            foreach (TradeOfferObject itemOffer in buyItems)
+            {
+                //Log("BOT", "For sale: " + itemOffer.Name);
+                if (itemOffer.Name == Task.Name)
+                    targetItem = itemOffer;
+            }
+
```
cyberjunk: I think this loop can be replaced. Just assign the return value of Data.Buy.Items.GetItemByName(Task.Name) to  'targetItem' above directly, instead of null.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -402,6 +678,190 @@ protected void DoUse(BotTaskUse Task)
...
+            if (finalAmount > 0)
+            {
+                // Buy the maximum or desired amount
+                targetItem.Count = finalAmount;
+
+                //ReqBuyItemsMessage buyMsg = new ReqBuyItemsMessage(Data.Buy.TradePartner.ID, new ObjectID[] { new ObjectID(targetItem.ID, finalAmount) });
+
+                ObjectID[] orderArray = new ObjectID[1];
+                orderArray[0] = new ObjectID(targetItem.ID, finalAmount);
+                SendReqBuyItemsMessage(Data.Buy.TradePartner.ID, orderArray);
```
cyberjunk: You should check Data.Buy.TradePartner somewhere at the beginning of this function for NULL.
If a server-save occurs after you have requested the offers, but not yet send your buy list, the DataController will call Invalidate(), which will also reset the Data.Buy.TradePartner to null. Also OgreClient and I think the original clients will abort a NPC buy interaction (npc offers window will close), when a server-save occurs.

Meridian59.Bot.Spell/SpellBotClient.cs
AlgorithmsOfMathdestruction: Okay, I haven't added anything outside the spellbot as I didn't keep the full project up to date. Got some experience with the git thing now. Thanks for pointing out where to add this.

Meridian59.Bot.Spell/SpellBotClient.cs
AlgorithmsOfMathdestruction: Yep, that's autolog. It caused the bot to crash probably caused by missing null tests. I just wanted to test that feature for training spells that needs to be casted in a fight area. Whatever .. seems to be a bad idea for a public release.

Meridian59.Bot.Spell/SpellBotClient.cs
AlgorithmsOfMathdestruction: Can do that, thought testing AvatarID when entering the task execution loop was enough. line 170/252

Meridian59.Bot.Spell/configuration.xml
AlgorithmsOfMathdestruction: Sry didn't notice. Just bashed the changes together yester afternoon to try a first commit and pull. Thought you did just higher the number of connection template. Going to fix that up :)

Meridian59.Bot.Spell/SpellBotClient.cs
AlgorithmsOfMathdestruction: For multiples use a const list to check for .Contains() ?

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -371,6 +539,115 @@ protected void DoCast(BotTaskCast Task)
...
         /// <summary>
+        /// Executes a Task 'eat'
+        /// </summary>
+        /// <param name="Task"></param>
+        protected void DoEat(BotTaskEat Task)
+        {
+
+            if (Task.Name.ToLower().Equals("feeding")) // eat as much as possible
+            {
+                int nutritToEat = 200 - Data.VigorPoints;
```
cyberjunk: I have added several 'shortcut' properties in the DataController instance for purposes like this:
>> int nutritToEat = 200 - Data.VigorPoints;

Data.Fatigue will tell you how much vigor you can get by resting.
Data.Hunger will tell you how much vigor you can get by eating.
Data.Injury will tell you how much hitpoints you can heal yourself.
Data.ManaMissing will tell you how much more mana you can restore.

Meridian59.Bot.Spell/SpellBotClient.cs
```diff
@@ -172,41 +254,93 @@ public override void Update()
...
-                if (currentTask == null)
+                //// Emergency log
+                //if (Data.HitPoints < (Data.AvatarCondition.GetItemByNum(1).ValueMaximum / 3))
+                //{
+                //    DoCast(new BotTaskCast("phase", "", "", "", 100));
+                //}
+
+                bool waitForRecover = false;
+                if (recoveringHp)
+                    waitForRecover = !ProcessIsStatRecovered("HP");
```
cyberjunk: The best place to track condition stat changes in HP, MP and VIG is within an overriden method of OnAvatarConditionListChanged().

Simply add to SpellBotClient.cs:
>> protected override void OnAvatarConditionListChanged(object sender, ListChangedEventArgs e)
{
   base.OnAvatarConditionListChanged(sender, e);
   // your code to handle changes in HP, MP, VIG here using Data.Hitpoints, Data.Fatigue etc. values
}

cyberjunk:
I really like the "recover" task here. May be you want to setup this as a dedicated pullrequest?
However there is few changes I would like to request compared to the approach here:

1) Use event handlers for ListChanged event of Data.AvatarCondition or PropertyChanged event of the included items there to run code when these stats change their values (to determine when recover-task should end).

2) I would prefer a XML notation like:
<task type="recover" hp="80" mp="60" vigor="80" />

The task will recover until all values are reached. Only by resting first - no eating. Values too high should be mapped to max value. Also 'max' should be a valid value for these xml attributes.

AlgorithmsOfMathdestruction:
Hey, just had job marathon before holidays ;) I think now I'll find the time to get back to implement stuff and create pullrequests.

I can do it the way you mentioned.

cyberjunk:
Closing this. Super old

PR #74, opened 2015-07-15
Spell bot attr cap

AlgorithmsOfMathdestruction:
Added "cap" attribute to task "cast".
Added XMLVALUE "self" ad value for "cast" Tasks "target" attribute.

-using selftarget forced "where" to room.
-leaving cap empty/0 will work like cap="99"
-cap above 99 will spin endless
-added .ToLower() on AMLVALUE check to ensure compability for diffenet write cases


cyberjunk:
I've made small modifications to the 'Cap' handling to this after merging it (and adjusted the example template a bit.)

Please have a look here:
https://github.com/cyberjunk/meridian59-dotnet/commit/98686313fae9f53eb18aac3621f380cf1254411f

PR #79, opened 2015-09-12
Refactor client gameplay preference options.

skittles1:
Attempt to add UC_REQ_PREFERENCES, UC_SEND_PREFERENCES and
UC_RECEIVE_PREFERENCES user commands added to server/classic client.

Currently UserCommandReceivePreferences doesn't work properly (it sets ClientPreferences to 0). Added PropertyChangedEventArgs for the newer commands but there are no UI elements or anything for them to call (not sure if they should be there yet).

I'm not sure any of this is correct, but hoping I'm on the right track.


Meridian59/Data/Models/UserCommand/UserCommandReceivePreferences.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            clientPreferences = BitConverter.ToInt32(Buffer, cursor);
+            cursor += TypeSizes.INT;
+         }
+
+         return cursor - StartIndex;
+      }
+      #endregion
+
+      public int clientPreferences { get; set; }
```
cyberjunk: Please use an uppercase starting letter for public properties
--> ClientPreferences

Same in the constructor parameter below.

Also please use the type "uint" for bitflags rather than "int".

Meridian59/Data/DataController.cs
```diff
@@ -464,8 +487,124 @@ public bool IsSafety
                 }
             }
         }
+#else
+           get { return ((ClientPreferences & CF_SAFETY_OFF) == 1); }
```
cyberjunk: Please make sure these calls look like:
(ClientPreferences & CF_SAFETY_OFF) == CF_SAFETY_OFF)

I think what you provide here would fail for:
ClientPreferences flags = 1010
CF_SOMETHING = 0010

The '&' would leave you with 0010 (dec:2) which is NOT equal 1, so it would return FALSE, but the flag was set.

Meridian59/Data/DataController.cs
```diff
@@ -59,6 +59,24 @@ public class DataController : INotifyPropertyChanged, ITickable
...
+
+        /// Flags for client preference options, e.g. safety.
+        public const int CF_SAFETY_OFF = 0x00000001; // Player has safety off
+        public const int CF_TEMPSAFE = 0x00000002; // Player has temp safety on death activated
+        public const int CF_GROUPING = 0x00000004; // Player is grouping
+        public const int CF_AUTOLOOT = 0x00000008; // Player is automatically picking up loot
+        public const int CF_AUTOCOMBINE = 0x00000010; // Player automatically combines spell items
+        public const int CF_REAGENTBAG = 0x00000020; // Player automatically puts items into reagent bag
+        public const int CF_SPELLPOWER = 0x00000040; // Player gets spellpower readout from cast spells
+#endif
```
cyberjunk: Please imlement the CF_ flags as a "PreferencesFlags.cs" class.

For an example have a look at *Files.ROO.BowtieFlags* or *Files.ROO.RooSideDefFlags*.
You can also look at ObjectFlags.cs, but it's huge and mentioned two files are more easy to understand.
Basically these classes derive from *Common.Flags* and provide easy get() and set() accessors for the values stored in them.

Your DataController class here then should have once instance of this *PreferencesFlags* class.

Meridian59.Ogre.Client/UIStatusBar.cpp
```diff
@@ -171,8 +175,12 @@ namespace Meridian59 { namespace Ogre
 
 	bool UICallbacks::StatusBar::OnSafetyClicked(const CEGUI::EventArgs& e)
 	{
+#if VANILLA
 		OgreClient::Singleton->SendUserCommandSafetyMessage(
 			!OgreClient::Singleton->Data->IsSafety);
+#else
+		OgreClient::Singleton->Data->IsSafety = !OgreClient::Singleton->Data->IsSafety;
```
cyberjunk: I see that you're setting the datalayer value here for OMP and then listen for the change in ControllerUI.cpp to send the message, while VANILLA directly send the message. I don't think this is ideal due to different approaches on OMP/VANILLA.

If the server does not already automatically echo back the preferences when they are changed and this is something you don't wand to add, then please send 2 messages in the OMP code here. First send your changed preferences, then request your preferences. This will automatically use the same code for updating the datalayer/UI as for receiving them anyways and will make sure everything is in sync.

Meridian59/Data/Models/PreferencesFlags.cs
```diff
@@ -0,0 +1,192 @@
...
+        private const uint CF_TEMPSAFE = 0x00000002; // Player has temp safety on death activated
+        private const uint CF_GROUPING = 0x00000004; // Player is grouping
+        private const uint CF_AUTOLOOT = 0x00000008; // Player is automatically picking up loot
+        private const uint CF_AUTOCOMBINE = 0x00000010; // Player automatically combines spell items
+        private const uint CF_REAGENTBAG = 0x00000020; // Player automatically puts items into reagent bag
+        private const uint CF_SPELLPOWER = 0x00000040; // Player gets spellpower readout from cast spells
+        #endregion
+
+        #region Constants
+        public new const string PROPNAME_FLAGS = "Flags";
```
cyberjunk: The base class (*Flags.cs*) already defines *PROPNAME_FLAGS*. Please use this instead instead of redefining it. Also make sure you're properties setter below raise the event at the end of the set. (Compare BowtieFlags.cs)

>> RaisePropertyChanged(new PropertyChangedEventArgs(PROPNAME_FLAGS));

Meridian59/Data/Models/PreferencesFlags.cs
```diff
@@ -0,0 +1,192 @@
...
+        /// Whether safety is enabled or not
+        /// </summary>
+        ///
+        public bool IsSafety
+        {
+           get { return (flags & CF_SAFETY_OFF) == CF_SAFETY_OFF; }
+           set
+           {
+              if (value) Value |= CF_SAFETY_OFF;
+              else Value &= ~CF_SAFETY_OFF;
```
cyberjunk: After this line and the others you need:
RaisePropertyChanged(new PropertyChangedEventArgs(PROPNAME_FLAGS));

Meridian59.Ogre.Client/UIStatusBar.cpp
```diff
@@ -23,15 +23,19 @@ namespace Meridian59 { namespace Ogre
...
-		
+
 		// attach listener to Data
 		OgreClient::Singleton->Data->PropertyChanged += 
 			gcnew PropertyChangedEventHandler(&ControllerUI::StatusBar::OnDataPropertyChanged);
-           
+
+		// attach listener to Preference Flags
+		OgreClient::Singleton->Data->ClientPreferences->PropertyChanged +=
+			gcnew PropertyChangedEventHandler(&ControllerUI::StatusBar::OnClientPreferencesChanged);
```
cyberjunk: I think this requires a "#if !VANILLA" check?
Same for the destroy method below...

Meridian59.Ogre.Client/UIStatusBar.cpp
```diff
@@ -171,8 +197,14 @@ namespace Meridian59 { namespace Ogre
 
 	bool UICallbacks::StatusBar::OnSafetyClicked(const CEGUI::EventArgs& e)
 	{
+#if VANILLA
 		OgreClient::Singleton->SendUserCommandSafetyMessage(
 			!OgreClient::Singleton->Data->IsSafety);
+#else
+		OgreClient::Singleton->Data->ClientPreferences->IsSafety
+			= !OgreClient::Singleton->Data->ClientPreferences->IsSafety;
```
cyberjunk: Now that the server echos back the preferences (?), I think you don't need flip this value manually here.

Meridian59/Client/BaseClient.cs
```diff
@@ -2662,12 +2666,14 @@ public void ExecChatCommand(string Text)
                         SendReqQuit();
                         break;
 #if !VANILLA
-                    case ChatCommandType.TempSafe:
-                        SendUserCommandTempSafe(((ChatCommandTempSafe)chatCommand).On);
+                    case ChatCommandType.TempSafe:
+                        Data.ClientPreferences.TempSafe = ((ChatCommandTempSafe)chatCommand).On;
```
cyberjunk: Same as other comment here. Probably not required to flip manually if preferences get sent back from server.
One more below.

PR #80, opened 2015-09-30
Add BP_STOP_WAVE protocol and infrastructure.

skittles1:
Stops a wav playing by resource ID sent from server.

Works correctly now.

### DIFF
--- a/Meridian59.Ogre.Client/ControllerSound.cpp
+++ b/Meridian59.Ogre.Client/ControllerSound.cpp
@@ -125,7 +125,7 @@ namespace Meridian59 { namespace Ogre
     };
 
 	void ControllerSound::UpdateListener(RemoteNode^ AvatarNode)
-    {
+	{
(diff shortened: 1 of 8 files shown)


Meridian59.Ogre.Client/ControllerSound.cpp
```diff
@@ -202,7 +206,69 @@ namespace Meridian59 { namespace Ogre
...
+
+      // get resource name of wav file
+      System::String^ sourcename = Message->PlayInfo->ResourceName->ToLower();
+
+      // native string
+      ::Ogre::String o_str = StringConvert::CLRToOgre(sourcename);
+      const char* c_str = o_str.c_str();
+
+      // check if sound is known to irrklang
+      ISoundSource* soundsrc = soundEngine->getSoundSource(c_str, false);
```
cyberjunk: I don'T think you need to require this anymore.

PR #105, opened 2016-01-23
fix for some dm commands being usable from chat

Daenks:
Im pretty sure this could be done better, but it needed to be done quick, so here it is. This code is running live.


Daenks:
included in #111

PR #107, opened 2016-01-24
Develop

GJiin:


Meridian59/Data/DataController.cs
```diff
@@ -873,14 +873,14 @@ public void Invalidate()
...
-            
+
             // clear single data models
             Effects.Clear(true);
             GuildInfo.Clear(true);
             GuildAskData.Clear(true);
             DiplomacyInfo.Clear(true);
             AdminInfo.Clear(true);
-            ClientPreferences.Clear(true);
+            ClientPreferences.Clear(false); // Do not propagate clear to server.
```
cyberjunk: This is the only changed line here right? Others seem formatting-related. What is it supposed to fix exactly?
I'm not sure if this is good, because Invalidate() is supposed to reset all local-data related to the currenty logged in character. This includes the per-toon saved preferences flags, which now would not get reset correctly.

GJiin: Yes, passed value is only change. What it fixes is that on server save,
user preferences are reset. The problem is that the preferences do get
reset and those reset states are sent to server. I believe only local state
should be reset so that server can overlay,

A Clear() to preferences was ADDED a little bit ago. This is what happens
as a result.

1) Server sends 'server save' command.
2) client begins a reset procedure
3) invalidate is called - various collections are Clear()ed. Including
preferences.
4) a preference.Clear(TRUE) resets all preferences to FALSE. ... and...
here its comes... wait for it..
5) preferences are propagated (sent) to server. NOTE: all -references are
FALSE!
6) the server turns around and sends the now all-false preferences.
7) client complies and sets all preferences to FALSE and as login, its
propagated to UI, etc.


Two things we can do:

1) completely remove (comment out) the preferences. Effectively reversing
the addition of the Clear() from skittles a few months ago.

or

2) pass a FALSE flag to Clear(), which resets preferences BUT DOES NOT
propagate the, now clear, flags to server.


Make sense?

-MK

cyberjunk: I'm sorry, I've had a quick review on the issue, and it seems there are only 2 main problems here:

1) The server-stored preferences are only requested from the server, when the user logs in the toon in the avatar-selection (This is in OgreClient.cpp, search: SendUserCommandReqPreferences) - on the other hand the preferences are reset (in the line you changed), on any server-save cycle. This obviously can't work well. First thing to figure out would be, whether the preferences should be considered stable across server-saves or whether they should be invalidated and re-aquired like the rest. First variant would mean not clearing them out, second one would mean requesting them with the other stuff, instead a trigger in the login-click.

2) There is an unwanted "binding-loopback". The clear() call will raise the propertychanged handlers in the OgreClient UIOptions, these will then send the "cleared" values. These cases (changed by user and changed by code) need to be differed.

GJiin: A request for server preferences is sent on each user change
A request for server preferences is sent on each server-save

the Clear(false) does reset preferences in the collection and as pointed
out to the UI,
Avoid the loop is exactly what is needed. That is done by passing false.

-mk

cyberjunk: >> Avoid the loop is exactly what is needed. That is done by passing false.

Tried to point out the "true" is required here to reset the UI. My suggestion is to move the actual user-created change into the "click" handler of the toggluebuttons, not the select changed.

Still I don't think the preferences are requested at serversave.
If they are supposed to, then there should be a call in BaseClient.cs "RequestInfoAfterServerSave()"

Still I don't see where

GJiin: Here's the problem.. at login and save, there is a call to get preferences
that originates from a key press which is not cleared. (This needs
addressing.) I did put a call to SendSendUserPreferences() in AfterSave...
 but found it unneeded. If the residual key trigger is fixed, yes, a call
to have server send preferences is needed.

BTW.. i dont feel resetting preference collection or UI is needed as that
will happen when server send preferences.

A good way to see is to run debug UI and watch network traffic at startup
and when you click the server save button.

[image: Inline image 1]

cyberjunk: Please have a look at:
https://github.com/cyberjunk/meridian59-dotnet/commit/4b8a3f90bcfdd20131a3b94c6031b1443063ef60

This seems to fix the issue for me, while still resetting the data so the UI resets to (=true param).
It first moves the "SendUserCommandReqPreferences()" into the corelibrary (BaseClient.cs - RequestInfoAfterLogin()" - This is also called on server-save cycles. This place is definitely the right place to do this, since any other client-implementation would need it too etc.

2) I resolved the issue when cleared preferences are sent to the server as result of the Clear() call running the same UI-handlers as the user-generated clicks by using the ClickEvent for user-generated clicks. This eventhandler is not also called by code-modifed togglebutton state (Clear)

PR #110, opened 2016-02-13
Add 104, 105 and 106 server profiles.

skittles1:
Added missing 104 profile, and the new 105/106 profiles to Ogre connection info and shop/spellbot config files.

### DIFF
--- a/Meridian59.Bot.Shop/configuration.xml
+++ b/Meridian59.Bot.Shop/configuration.xml
@@ -85,6 +85,32 @@
       </ignorelist>
     </connection>
     <connection
+      name="105"
+      host="meridian105.meridiannext.com"
+      port="5905"
+      useipv6="false"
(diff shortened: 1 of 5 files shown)


cyberjunk:
I merged the new RSB files on the other repo, but this nested git (meridian59-resources) in subfolder /resources won't get updated automatically, it requires a commit to change the link from the mainrepo to the new head of the child repo

cyberjunk:
PS: See the last commit about chaning the linked resources repo:

https://github.com/cyberjunk/meridian59-dotnet/commit/56bdd7b1233036a98254d53e2a81652fd13efd39

cyberjunk:
PPS:
I'm not sure how to deal with the submodule repositories exactly in case you need to change both.
I think I have to merge the pullrequest(s) on the sub-repositories first. Then you can add the commit to the pullrequest of the main-repository, which updates the linked in submodules.

PR #111, opened 2016-02-16
Add ability for IRC bots to echo each other's output to their own server.

skittles1:
This includes #105 as both PRs are being used in production and edit the same code. I can rebase etc. if necessary or make modifications here.

##### Draft of inter-server communication using IRC bots.

Two IRC bots in the same IRC channel (and on different servers) will be able to output the broadcasts and deaths from one server to another. E.g. if Server103Chat echoes a broadcast from server 103, Server105Chat can convert it back to a m59-compatible string and broadcast it on 105, using the dm gqemote command to avoid feedback loops.

For a bot to echo other bots' text, it will need to list that bot's name and a banner (for when the text hits the bots own server) in configuration.xml.

### DIFF
--- a/Meridian59.Bot.IRC/IRCChatStyle.cs
+++ b/Meridian59.Bot.IRC/IRCChatStyle.cs
@@ -21,6 +21,47 @@ using System;
 namespace Meridian59.Bot.IRC
 {
     /// <summary>
+    /// Extensions for the String class that allow replacing the first instance
+    /// of a substring in a string, starting from pos 0 or given pos.
+    /// </summary>
+    public static class StringExtensionMethods
+    {
+        /// <summary>
+        /// Replaces the first instance of a search string in the string with another string.
+        /// </summary>
+        /// <param name="search"></param>
+        /// <param name="replace"></param>
+        /// <returns></returns>
+        public static string ReplaceFirst(this string text, string search, string replace)
+        {
+            int pos = text.IndexOf(search);
+            if (pos < 0)
+            {
+                return text;
+            }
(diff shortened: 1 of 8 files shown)


Meridian59.Bot.IRC/IRCChatStyle.cs
```diff
@@ -21,6 +21,47 @@
 namespace Meridian59.Bot.IRC
 {
     /// <summary>
+    /// Extensions for the String class that allow replacing the first instance
+    /// of a substring in a string, starting from pos 0 or given pos.
+    /// </summary>
+    public static class StringExtensionMethods
```
cyberjunk: Any chance you can add the methods included here to the core lib?
Preferrably into Meridian59.Common namespace..

I think it's very generic and probably not limited to the IRC-Bot....
Also please remove the { } in the (pos < 0) cases and use upper-case parameter names.

I know that the extension-technique seems nice in C# as it seamlessly integrates into the class that gets extended.
That's not fully working in C++ CLI however - you can call these but ont from an object reference like you can do in C#.

The class Meridian59.Common.Util already contains some String util functions, such as "CountSubStringOccurences()" for example. But I decided to not use the extension-technique, due to described limitations to C# above.

cyberjunk:
I don't really have time to maintain all the small tools...
Since you're the ones who actually use IRC-Bot, you can basically do whatever you want with it... :)

Means: Let me know when to merge this...

I'll try to review parts here, especially if they change the core-lib.
But probably I won't inspect the IRC-Bot related changes much.

skittles1:
I think it's ready for merge, we've had it running for a couple weeks now with no issues

PR #113, opened 2016-02-19
Add any missing servers to dropdown/configuration.

skittles1:
After loading the connection profiles from configuration.xml, perform a check for each server and add it if missing from config file.

This works for me locally and I'm able to log into the new profiles, but not sure if this is the correct solution or if there's something better.

### DIFF
--- a/Meridian59/Common/Config.cs
+++ b/Meridian59/Common/Config.cs
@@ -478,7 +478,25 @@ namespace Meridian59.Common
                         password,
                         character,
                         ignorelist));
-                }               
+                }
+                // Double check we have all available servers in connections list.
+#if !VANILLA
+                if (!HasConnection(ConnectionInfo.CON103.Host, ConnectionInfo.CON103.Port))
+                    connections.Add(ConnectionInfo.CON103);
+                if (!HasConnection(ConnectionInfo.CON104.Host, ConnectionInfo.CON104.Port))
+                    connections.Add(ConnectionInfo.CON104);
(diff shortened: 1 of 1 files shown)


skittles1:
Wondering if we should leave 104/106 out of the list as they're both test servers, so we don't confuse any users who might otherwise log on and see no players.

cyberjunk:
>> Wondering if we should leave 104/106 out of the list as they're both test servers, so we don't confuse any users who might otherwise log on and see no players.

I leave that decision up to you. Pull looks fine otherwise.

skittles1:
Thinking probably best to keep them now - the benefit of connecting with Ogre to the test servers probably outweighs any issues with accidental connections.

Meridian59/Common/Config.cs
```diff
@@ -16,6 +16,7 @@
 
 using System;
 using System.IO;
+using System.Linq;
```
cyberjunk: I have not used Linq at any other place in the corelib so far...

I think the reason was it being quite buggy on other platforms than Microsoft .NET (like MonoDroid etc.)
That's however couple of years ago and might not be true anymore.

Still, while the linq call seems handy, please rather add a method to Config.cs like "HasConnection(...)" or something to check the connections collection for the existance of an item. That's the way things are implemneted in basically any collection code in the corelib...

skittles1:
Added the HasConnection() method to iterate through the BindingList, I think this is what you meant? BindingList has a Contains method but I'm not sure if i should be using that with the default connection profiles rather than checking specifically for Host and Port.

cyberjunk:
Yes exactly, thx

PR #114, opened 2016-02-19
Modify Ogre client to use OGG format for sound/music.

skittles1:
Ogre will now convert string names for music resources (which may have extension .mp3/.wav/.mid) to .ogg. I'm not sure what else I need to change here, as this enables both sound and music to play on 104 (resources .wav/.mp3) and 106 (resources .ogg).

Also not sure which is the best way for me to submit the .ogg files - should I make a separate PR to the meridian59-resources repo, or upload the files somewhere for you to add them?

### DIFF
--- a/Meridian59/Common/Constants/ResourceStrings.cs
+++ b/Meridian59/Common/Constants/ResourceStrings.cs
@@ -15,6 +15,7 @@
 */
 
 using Meridian59.Common.Enums;
+using System.IO;
 using System.Collections.Generic;
 
 namespace Meridian59.Common.Constants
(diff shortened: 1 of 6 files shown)


Meridian59/Common/Constants/ResourceStrings.cs
```diff
@@ -365,18 +366,18 @@ public static bool Is(string Value)
...
+            public const string OUCHMALE4 = "ouchm4.ogg";
+            public const string OUCHFEMALE1 = "ouchf1.ogg";
+            public const string OUCHFEMALE2 = "ouchf2.ogg";
+            public const string OUCHFEMALE3 = "ouchf3.ogg";
+            public const string OUCHFEMALE4 = "ouchf4.ogg";
 
             public static HealthStatus IsOuch(string Value)
             {
-                switch (Value)
+                switch (Path.ChangeExtension(Value, ".ogg"))
```
cyberjunk: I guess you could use the constant "FileExtensions.OGG" in all these calls instead of ".ogg" ?

cyberjunk:
The code changes look basically good to me.
However the meridian59-resources repository needs to have all its WAV and MP3 files replaced by the new OGG variants.

cyberjunk:
I can merge this as soon as you setup a second pullrequest on the resources repository (https://github.com/cyberjunk/meridian59-resources) which replaces all WAV and MP3 with OGG files.

After I merged both I can make the commit which updates the linking myself.

PR #115, opened 2016-02-21
Add server-specific .roo file loading.

skittles1:
There are now 3 OpenMeridian servers open and due to differing update
schedules and content, the same .roo file may be different on each
server. Currently this is the case for Raza, where the version of the
city (7 rooms anyway) is different on 103/112 and 105.

To allow the single Ogre client to work on all 3 servers, I have changed
.roo loading to first check a server-specific folder (currently this is
the server name/number) for any .roo files before checking the parent
rooms directory. 103 would have its changed rooms in \rooms\103\, 105 in
\rooms\105\ etc.

Specifically preloading rooms on client startup has been removed
(default .roos are still loaded in Init) and when a user tries to
connect to a server, the additional .roos are loaded and if preloading
rooms is enabled, this occurs in the same thread. The server's
name/number is saved as RoomsSubFolder for future use on room loads, and
for removing the extra rooms from the Rooms dictionary if the user
switches between servers.

Tested this using 106/112 and the Raza rooms, with various combinations
of preloading/no preloading, empty folders, extra roo files in one folder,
switching between servers with quit.

### DIFF
--- a/Meridian59.Ogre.Client/OgreClient.cpp
+++ b/Meridian59.Ogre.Client/OgreClient.cpp
@@ -654,7 +654,6 @@ namespace Meridian59 { namespace Ogre
 		ResourceManager->Preload(
 			Config->PreloadObjects,
 			Config->PreloadRoomTextures,
-			Config->PreloadRooms,
 			Config->PreloadSound,
 			Config->PreloadMusic);
 
--- a/Meridian59/Client/BaseClient.cs
+++ b/Meridian59/Client/BaseClient.cs
@@ -116,6 +116,9 @@ namespace Meridian59.Client
                 Config.SelectedConnectionInfo.StringDictionary,
 				LanguageCode.English); // todo: from config
 
+            // Load server-specific rooms if needed
+            ResourceManager.AddServerRooms(Config.SelectedConnectionInfo.Name, Config.PreloadRooms);
+
 		    // fill ignore list in datacontroller with ignored
             // playernames for this connectionentry.
 		    Data.IgnoreList.Clear();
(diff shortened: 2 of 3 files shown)


skittles1:
Curious what you think about this one... server 105 is live with changed rooms but we forgot to think what would happen when Ogre connected :)

cyberjunk:
I didn't had time to review this in detail.
There are two things I'm not perfectly happy about

(1) Adding server-dependency on the resources other than the string-dicionaries makes it hard to preload them. As you noticed in this change this can't be done where it is done right now. Technically it could be done after the connect button is pressed and before the avatar is selected, or preferrably (this is what I had in mind once), fully in background until you try to login an actual toon (which would then force you to wait if preloading is enabled). Unfortunately this is no trivial change and therefore rather out of scope...

(2) Changing the resources so only one kind of them is now server-dependant with its own subfolders is kinda introducting two patterns where one would be enough. I see why there are reasons to have server-dependant resources. If we need this, we should probably migrate all resource types to this pattern.

cyberjunk:
A possible approach could be a two-step solution where:

1) Switches all resources to a new server-dependant subfolder structure. I guess this is pretty much what you did here with rooms. So a room should be store in like /rooms/103/someroom.roo and a sound in /sounds/103/somesound.wav. If there is huge amounts of identical resources, we could think about a folder "/common/" which will be tried to looked up, in case the specific room folder did not have an entry (nice but more work). This step would require to disable all preloading of legacy resources. Please keep in mind that there are additional ogre related resources, like 3D models with their own structure, which if correcly migrated to a new server-dependant resource pattern, would require adjustements as well.

2) Re-add preloading capabilities. Either with a loading-bar after connect or more in background.

skittles1:
The way I've written it, /rooms/ is the common one and is checked after the server-specific one (so this could change to /rooms/common/). Most resources would be only common, especially stuff like sounds/room textures.

I can try and convert all the legacy resources to load after connect (in background threads?) if that's the way to go.

skittles1:
Implemented for objects/roomtextures/rooms/sounds/music in #117

PR #116, opened 2016-02-22
Prevent room grid loading on roo version >= 15.

skittles1:
Version 15 rooms no longer include the server movement flags/grids, so don't try to load them.

### DIFF
--- a/Meridian59/Files/ROO/RooGrids.cs
+++ b/Meridian59/Files/ROO/RooGrids.cs
@@ -44,6 +44,9 @@ namespace Meridian59.Files.ROO
         {
             get 
             {
+                if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                    return 0;
+
                 // rows + cols
                 int len = TypeSizes.INT + TypeSizes.INT;
 
@@ -71,6 +74,9 @@ namespace Meridian59.Files.ROO
 
         public virtual int WriteTo(byte[] Buffer, int StartIndex = 0)
         {
+            if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                return 0;
+
             int cursor = StartIndex;
 
             Array.Copy(BitConverter.GetBytes(Rows), 0, Buffer, cursor, TypeSizes.INT);
@@ -136,6 +142,9 @@ namespace Meridian59.Files.ROO
 
         public virtual unsafe void WriteTo(ref byte* Buffer)
         {
+            if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                return;
+
             *((int*)Buffer) = Rows;
             Buffer += TypeSizes.INT;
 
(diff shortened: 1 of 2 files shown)


Meridian59/Files/ROO/RooGrids.cs
```diff
@@ -71,6 +74,9 @@ public virtual int ByteLength
 
         public virtual int WriteTo(byte[] Buffer, int StartIndex = 0)
         {
+            if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                return StartIndex;
```
cyberjunk: This function is supposed to return the overall count of written bytes.

See the lastline in each of these:
>> return cursor - StartIndex;

So your return value should be 0 here.

Meridian59/Files/ROO/RooGrids.cs
```diff
@@ -197,6 +206,9 @@ public virtual unsafe void WriteTo(ref byte* Buffer)
 
         public virtual int ReadFrom(byte[] Buffer, int StartIndex = 0)
         {
+            if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                return StartIndex;
```
cyberjunk: Basically the same as with the WriteTo() comment.
Should return the overall amount of bytes read here.
That would be 0...

Meridian59/Files/ROO/RooGrids.cs
```diff
@@ -269,6 +281,9 @@ public virtual int ReadFrom(byte[] Buffer, int StartIndex = 0)
 
         public virtual unsafe void ReadFrom(ref byte* Buffer)
         {
+            if (RooVersion >= RooFile.VERSIONNOGRIDS)
+                return;
```
cyberjunk: The unmanaged/pointer variants both look OK. No need to change anything here.
These are supposed to increment the pointer argument according to what they read from the pointer.
In this case nothing to read, nothing to modify.

skittles1: If i take out lines 284/285 and try to load a room without the grids it attempts to read it in and crashes.

cyberjunk: I'm afraid my sentence was easy to misunderstand.
What I mean: Your change looks totally perfectly OK! Nothing to change (on top of what you propose) here.
Removing this will crash of course :)

cyberjunk:
Thanks for doing this. Changes look good now.
Any chance you can verify that RooViewer is still able to load a vanilla room from 101?
I guess you tested startup in OgreClient...

skittles1:
Tested roo version 14/15 in Ogre client, will try RooViewer now.

skittles1:
Success, RooViewer will open all types of rooms (14, 15, vanilla).

cyberjunk:
Ok. Merged :-)

PR #119, opened 2016-03-12
Add LightingInfo and LightFlags classes, add highlighting light effect used on signs.

skittles1:
#### Commit 1
Change LightFlags from ushort to its own type, add LightingInfo class to hold the new LightFlags plus the light intensity and color (still byte/ushort respectively).

#### Commit 2
Replicate classic client's highlight effect for signs by lowering the range on these lights to 12% of normal.

### DIFF
--- a/Meridian59.AdminUI/Viewers/InventoryObjectView.Designer.cs
+++ b/Meridian59.AdminUI/Viewers/InventoryObjectView.Designer.cs
@@ -55,9 +55,7 @@
             this.colOverlayRID = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.colNameRID = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.colFlags = new System.Windows.Forms.DataGridViewTextBoxColumn();
-            this.colLightFlags = new System.Windows.Forms.DataGridViewTextBoxColumn();
-            this.colLightIntensity = new System.Windows.Forms.DataGridViewTextBoxColumn();
-            this.colLightColor = new System.Windows.Forms.DataGridViewTextBoxColumn();
+            this.colLightingInfo = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.colColorTranslation = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.colEffect = new System.Windows.Forms.DataGridViewTextBoxColumn();
             this.colAnimation = new System.Windows.Forms.DataGridViewTextBoxColumn();
@@ -110,9 +108,7 @@
             this.colOverlayRID,
             this.colNameRID,
             this.colFlags,
-            this.colLightFlags,
-            this.colLightIntensity,
-            this.colLightColor,
+            this.colLightingInfo,
             this.colColorTranslation,
             this.colEffect,
             this.colAnimation,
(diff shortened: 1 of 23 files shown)


cyberjunk:
I like the idea, it also seems good how you implemented it.

>> One issue I wasn't sure of was the flags, I've hardcoded the 0x08 because I didn't know if I should add a LightFlags enum or bitfield and existing references to the light flags just use a number. Let me know what your preference is there and I'll add it to this PR.

I think the only place I really evaluate the LightFlags value so far is in OgreClient project in RemoteNode.cpp the OnRoomObjectPropertyChanged(). But this will simply check the ushort LightFlags for greater 0 to decider wether to add a light or not. Currently there is (was) only one light-type, the typical point light. Not sure what you want to do with this, if you're going to put more encoded enums or values into these 16 bits, creating a wrapper-class in Meridian59.Data.Models like with all the XYFlags.cs files there would probably be nice.

skittles1:
It's prob fine like this for now, if more lighting effects get added it won't be for quite a while.

skittles1:
BTW this is in the live client if you want to merge it, had to release it as the sign highlighting flags were live and the rooms are flooded with light without it.

cyberjunk:
I really don't want to add the hardcoded:
>> & 0x08

Can you point out where the 16 bits for the lightflags are specified/built on the server-side?
I will try to add a "LightFlags" class to Meridian59.Data.Models just like for all the other Flags.

skittles1:
Sorry just saw this... it's built in several places depending on what sends the light source - 2 bytes for flags, 1 byte for intensity and 2 bytes for color.

The defines for flags are in proto.h line 605:
```
#define LIGHT_FLAG_NONE		0x0000
#define LIGHT_FLAG_ON		0x0001
#define LIGHT_FLAG_DYNAMIC	0x0002
#define LIGHT_FLAG_WAVERING	0x0004
#define LIGHT_FLAG_HIGHLIGHT 0x0008
```
LIGHT_FLAG_WAVERING isn't actually used for anything as far as I can tell. I tried to implement these as flags but had some issues with the types (ushort/bitwise operations), I could try it again if you don't have the time to implement it.

cyberjunk:
>> I tried to implement these as flags but had some issues with the types (ushort/bitwise operations),

Use "uint" as datatype when you want to do lots of bit operations.
There is lots of examples of Flag classes, which all internally use uint.

skittles1:
@cyberjunk I've added a new LightFlags class to remove the hardcoded 0x08, and also added a LightingInfo class to hold all three light-related fields (flags, intensity, color). I'm not sure if you wanted to take it to that level of abstraction, so it's fine if you want me to remove that and just use the LightFlags class in place of the ushort LightFlags value.

Works with dynamic/static/highlight lights, projectiles, lighting change (switching lights on and off).

cyberjunk:
Very nice, Delerium.
For the UI there is two solutions:
I'd be OK with both.

**1) Quick (keeps info in table row)**

a) Remove two of the 3 light columns like [here](https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.AdminUI/Viewers/RoomObjectsView.Designer.cs#L449) in the RoomObjects Grid.

b) Adjust the remaining one so that its _DataPropertyName_ matches the new LightInfo sub instance property name.

c) Overwrite the ToString() method in _LightInfo_ so that is returns the values for flags, intensity etc written to a string. The grid-cell will use the ToString() method for contents by default. That is also how the Flags are done.

**2) Advanced (moves data to subview)**
Well, basically like it's done with "Animation" and other sub instances, that are shown for the selected entry only.. You gotta create a new UI, set the model as datasource etc.

skittles1:
So this is what I have with option 1, mirroring how Position3D is handled:

<p align="center">
   
</p>

Not sure if the lighting info is important enough to go into its own subview.

Was there anything you wanted me to change in the 1st commit?

cyberjunk:
There is a bug in this it seems...
The buy-window is broken now, more correctly the parser throws an exception.

cyberjunk:
Can you create a new pullrequest for this?
I've reset the merge on the master.

Setting a breakpoint [here](https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Protocol/GameMessages/GameMode/BuyListMessage.cs#L67) might be start for debug. It doesn't seem able to parse this object.

Also there is an exception in Meridian59.log with the packetdump.

skittles1:
Found the issue, sorry will submit a fixed branch soon.

So would LightingInfo have a uint field Flags then, and be queried with LightingInfo.IsLightOn etc, since LightingInfo could have the IsLightOn method and check flags itself?

cyberjunk:
>> So would LightingInfo have a uint field Flags then, and be queried with LightingInfo.IsLightOn etc, since LightingInfo could have the IsLightOn method and check flags itself?

Yes. Basically just move everything from _LightFlags_ into _LightInfo_.
The protected uint field, as well as the public properties etc.

PR #122, opened 2016-04-11
In-client update process for Ogre client (work in progress).

skittles1:
This isn't quite finished yet - it works (can patch Ogre from 106 using this), but most of the code is in BaseClient.cs when it should probably be elsewhere and there's some hardcoded strings that need to change. Making the PR early because I need a bit of guidance on how you would like this code organised.

I think the message from the server (ClientPatchMessage.cs) and the info itself (ClientPatchInfo.cs) are probably okay, but I'm unsure about where the classes for the JSON patch file (patchinfo.txt) should be (the classes PatchFile and PatchFileList at the end of BaseClient.cs).

Also not sure if the way I'm handling the download form (ClientPatchForm.cs) or WebClient are ideal.

### DIFF
--- a/Meridian59.Ogre.Client/OgreClient.cpp
+++ b/Meridian59.Ogre.Client/OgreClient.cpp
@@ -824,17 +824,6 @@ namespace Meridian59 { namespace Ogre
 		Disconnect();
 	};
 
-   void OgreClient::HandleClientPatchMessage(ClientPatchMessage^ Message)
-   {
-      // Disconnect from server.
-      Disconnect();
-      // Set UI mode to download, start download bar.
-      Data->UIMode = UIMode::Download;
-      ControllerUI::DownloadBar::Start();
-      // Download the client patch cache and updater, launch it and disconnect.
-      DownloadHandler->DownloadClientPatch(Message->ClientPatchInfo);
-   };
-
 	void OgreClient::HandleDownloadMessage(DownloadMessage^ Message)
 	{
 		SingletonClient::HandleDownloadMessage(Message);
--- a/Meridian59.Ogre.Client/OgreClient.cpp
+++ b/Meridian59.Ogre.Client/OgreClient.cpp
@@ -826,10 +826,13 @@ namespace Meridian59 { namespace Ogre
 
    void OgreClient::HandleClientPatchMessage(ClientPatchMessage^ Message)
    {
-      // tell user about mismatching major/minor version
-      ::System::Windows::Forms::MessageBox::Show(APPVERSIONMISMATCH);
-
+      // Disconnect from server.
       Disconnect();
+      // Set UI mode to download, start download bar.
+      Data->UIMode = UIMode::Download;
+      ControllerUI::DownloadBar::Start();
+      // Download the client patch cache and updater, launch it and disconnect.
+      DownloadHandler->DownloadClientPatch(Message->ClientPatchInfo);
    };
 
 	void OgreClient::HandleDownloadMessage(DownloadMessage^ Message)
--- a/Meridian59/Meridian59.csproj
+++ b/Meridian59/Meridian59.csproj
@@ -115,6 +115,7 @@
     <Compile Include="Common\Events\InventoryObjectEvent.cs" />
     <Compile Include="Common\Events\ActionTypeEvent.cs" />
     <Compile Include="Common\Events\AvatarActionEvent.cs" />
+    <Compile Include="Common\Events\IntegerEvent.cs" />
     <Compile Include="Common\Events\StringEvent.cs" />
     <Compile Include="Common\Flags.cs" />
     <Compile Include="Common\GameTick.cs" />
@@ -555,6 +556,7 @@
     <Compile Include="Protocol\MessageController\MessageControllerClient.cs" />
     <Compile Include="Protocol\MessageController\MessageControllerHook.cs" />
     <Compile Include="Protocol\MessageController\MessageParser.cs" />
+    <Compile Include="Protocol\DownloadHandler.cs" />
     <Compile Include="Protocol\SubMessage\SubMessage.cs" />
     <Compile Include="Protocol\SubMessage\SubMessageGeneric.cs" />
     <Compile Include="Protocol\SubMessage\SubMessageNewCharInfo.cs" />
--- a/Meridian59/Protocol/DownloadHandler.cs
+++ b/Meridian59/Protocol/DownloadHandler.cs
@@ -47,6 +47,7 @@ namespace Meridian59.Protocol
         protected const string SUCCESSMESSAGE = "Success!";
         protected const string DOWNLOADFAIL = "Failed to download file.";
         protected const string CACHELOADFAIL = "Unable to load patch file. Please contact an administrator.";
+        protected const long NUMFILEBYTES = 1024;
         #endregion
 
         #region Event Handlers
@@ -235,9 +236,9 @@ namespace Meridian59.Protocol
             }
 
             // Number of bytes to read.
-            long numBytesToRead = 1024;
+            long numBytesToRead = NUMFILEBYTES;
             // Make sure we dont read past the end of a file smaller than 1024 bytes.
-            if (stream.Length < 1024)
+            if (stream.Length < NUMFILEBYTES)
                 numBytesToRead = stream.Length;
 
             byte[] fileBytes = new byte[numBytesToRead];
--- a/Meridian59/Protocol/DownloadHandler.cs
+++ b/Meridian59/Protocol/DownloadHandler.cs
@@ -1,0 +1,287 @@
+﻿/*
+ Copyright (c) 2012-2013 Clint Banzhaf
+ This file is part of "Meridian59 .NET".
+
+ "Meridian59 .NET" is free software: 
+ You can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation, 
+ either version 3 of the License, or (at your option) any later version.
+
+ "Meridian59 .NET" is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY;
+ without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ See the GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License along with "Meridian59 .NET".
+ If not, see http://www.gnu.org/licenses/.
+*/
+
+using System;
+using System.IO;
+using System.ComponentModel;
+using System.Collections.Generic;
+
+using Meridian59.Common;
+using Meridian59.Common.Enums;
+using Meridian59.Common.Events;
+using Meridian59.Data.Models;
+using Meridian59.Protocol.GameMessages;
+using Meridian59.Protocol.Enums;
+using Meridian59.Protocol.Events;
+
+using System.Net;
+using System.Drawing;
+using System.Linq;
+using System.Text;
+using Newtonsoft.Json;
+using System.Diagnostics;
+
+namespace Meridian59.Protocol
+{
+    /// <summary>
+    /// Handles the downloading of client patch files and updater launch.
+    /// </summary>
+    public class DownloadHandler : IDisposable
+    {
+        #region Constants
+        protected const string DEFAULTUPDATEFILE = "club.exe";
+        protected const string DEFAULTUPDATEFILEPATH = "..\\";
+        protected const string SUCCESSMESSAGE = "Success!";
+        protected const string DOWNLOADFAIL = "Failed to download file.";
+        protected const string CACHELOADFAIL = "Unable to load patch file. Please contact an administrator.";
+        #endregion
+
+        #region Event Handlers
+        public event EventHandler DownloadStarted;
+        public event EventHandler<StringEventArgs> DownloadText;
+        public event EventHandler<IntegerEventArgs> DownloadProgress;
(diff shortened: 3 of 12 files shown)


Meridian59/Meridian59.csproj
```diff
@@ -84,6 +84,7 @@
     <Reference Include="System" />
     <Reference Include="System.Drawing" />
     <Reference Include="System.ServiceProcess" />
+    <Reference Include="System.Windows.Forms" />
```
cyberjunk: Please don't add a reference to "System.Windows.Forms" to the corelib.
This is not platform independent but the corelib is supposed to be.

cyberjunk:
Had a quick review. The first commit with the new datamodel looks pretty much fine.

I'm not exactly sure about the VANILLA thing, may be some parts of this should be disabled for VANILLA build (but can always do this later).

Way more interesting is the question about the UI and the second commit.
I kinda thought this update-mechanism would not require an own dedicated UI...
I assumed this is all done by club.exe?

So far I have kept the UI out of the corelib, because System.Windows.Forms is not platform independent, won't run on Monodroid for Android for example.

If you absolutely need a Windows.Forms UI, then add a another c# library project to the solution (may be name it "UpdaterUI". Put your Windows.Forms elements in there and use them from the OgreClient. This is basically how like the AdminUI is integrated. Another option would be building the Update-UI using OgreClient's CEGUI main UI.

skittles1:
VANILLA shouldn't have an issue (nor should 103) as they won't send the protocol to trigger this (except if they want to take advantage of it specifically).

I probably should have realised the UI separation, it doesn't technically need a lot of UI except downloading patchinfo.txt and (potentially) club.exe might take 5-20 sec so I had a dialog box/form pop up to show the progress. BTW I have it working this way on classic so that club.exe is always checked/updated before meridian.exe calls it (so each exe will update the other).

My preferred solution would be a nicer-looking OgreClient UI for the download which I likely lack the skill to make :) I could give it a try though, and if it doesn't work out then try the UpdateUI project method? I was also thinking about the ability of spellbot/shopbot/ircbot to use the form, though on second thought those clients could just display it in text (it will be fairly easy to add support for these to club/blakserv in the future).

cyberjunk:
>> I was also thinking about the ability of spellbot/shopbot/ircbot to use the form, though on second thought those clients could just display it in text

Yes, I thought about that too. But the updates should be handled UI less / text only for the bots.
Because they can be run as Windows-Service without any UI/user interaction.

So there is no problem in making a nice update-UI with CEGUI inside OgreClient

cyberjunk:
>> My preferred solution would be a nicer-looking OgreClient UI for the download which I likely lack the skill to make :) I could give it a try though

I'm sure you can get this done :)
Quick:

1) Copy & paste good one in ControllerUI.h (like LoadingBar class in there)
2) Copy & paste the same .cpp file for the header-defs you copy&pasted in (1)
3) Rip out stuff you don't need / make it a bare skeleton
4) Init your new ui-sub element in ControllerUI::Initialize() and ControllerUI::Destroy() like all the other UI sub elements
5) Define the visiblity in ControllerUI::OnDataPropertyChanged()
6) You possibly need a need UIMode for this (-> Common.Enums.UIModE)
...

skittles1:
Seems easy enough, though is it an issue to put EventHandlers in BaseClient.cs? Since following LoadingBar example, DownloadBar will need those for updating filename, progress and success/errors.

skittles1:
Okay rewrote the 2nd commit - the layout for the Ogre UI is in the resource submodule obviously but otherwise Ogre UI is working fine. I moved all the patch code into a new file, DownloadHandler.cs to get the event handlers out of BaseClient.cs. That part is much the same as it was before, except no UI elements.

I'm not sure if there is a proper way to close the client, or if it should just call some form of exit(0); when club.exe runs.

cyberjunk:
>> I'm not sure if there is a proper way to close the client..

To exit just flip the 'IsRunning' property of RootClient.cs (or any other deriving class) to 'True'.
This will make exit the MainLoop and call the "Cleanup()" method (see RootClient.cs - method Start()).

In Application.cpp in OgreClient you can find the call to Start() in the WinMain() function there.
Once the mentioned flag is flipped, the code there will reach the "return 0"

skittles1:
Thanks... couple more questions:

With the layout, should I submit that in a separate PR to the meridian59-resources repo like I did with the sounds? I'm still not exactly sure of the best way to submit PRs to submodules in a way that links them to the main one (and the merge commit there could render any resources commit I make here out of date). FYI this is what it looks like in the debug client: http://i.imgur.com/tNluhKc.png

I used the Newtonsoft JSON library for deserializing patchinfo.txt but now Ogre requires Newtonsoft.Json.dll to run, should I look for a way to do this that doesn't require the extra file? I suspect all the clients now likely require this but haven't attempted building them yet, though I do want to get the updater set up for the others pretty soon anyway. 

Last one, I'm not sure if I need the HandleClientPatch stub in MessageController.cs as I don't think anything needs to be done there.

I think there's some variables/spacing that still need fixing up and I'll do that tonight along with adding the client exit, please let me know if anything else needs changing.

cyberjunk:
The image looks nice :)

>> With the layout, should I submit that in a separate PR to the meridian59-resources repo like I did with the sounds? I'm still not exactly sure of the best way to submit PRs to submodules in a way that links them to the main one

That's a good question...
Locally I work with the submodules like this:

1) Commit to the submodule (executed on sub-folder 'Engine' or 'Resources')
2) Push to the submodule (executed on sub-folder 'Engine' or 'Resources')
3) Commit to mainrepo (shows changed submodule revision link as change)
4) Push to mainrepo

No idea if this can be combined somehow in github pullrequests easily, because it contains two pushes.
And the second can only be done once the first is done due to the additional commit.

For github this would may be mean:
1) Make a PR for the updated resources repo
2) Make a PR for the updated mainrepo
3) Wait for me to accept (1)
4) Add a commit to PR from (2) updating the submodule revision
5) Wait for me to accept (2)

Meridian59/Meridian59.csproj
```diff
@@ -81,7 +81,12 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Newtonsoft.Json, Version=4.5.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\Tools\ircdotnet\lib\Samples\Newtonsoft.Json.dll</HintPath>
```
cyberjunk: This looks like the Json dll is linked from the ircbot subfolder into the core-lib?..

cyberjunk:
>> I used the Newtonsoft JSON library for deserializing patchinfo.txt but now Ogre requires Newtonsoft.Json.dll to run...

It seems you added a library reference to the core-lib.
So anything linking Meridian59.dll will also require Newtonsoft.JSON.dll

I'm not a big fan of additional dependencies on the core-lib due to portability (unless it's absoluteltey needed). Not sure about the portability of Newtonsoft JSON library.

I have to admin I never used JSON in .net - but I really kinda expected there are some internal framework classes to deal with JSON nicely.

If it's impossible without Newtonsoft library, then please check how compatible this is with MonoDroid and Mono. Otherwise consider doing something I did with the "DRAWING" symbol in the core-library (See project settings). This symbol is used to disable some classes from Meridian59.Drawing namespace (like ImageComposerGDI.cs) in the "Any-CPU" (less features but runs on non-windows) build of the corelib.

cyberjunk:
>> Last one, I'm not sure if I need the HandleClientPatch stub in MessageController.cs as I don't think anything needs to be done there.

Unless you have some content for that stub, you don't need that...
Basically the MessageController runs in the network thread and it has to perform the following steps:

1) Create a typed instance of the GameMessage class (parses data from bytes)
--> You are doing this

2) Resolve the resource-strings from the IDs (= Execute the function "ResolveStrings()" for instances of IStringResolvableance in the models you parsed in (1))
--> Unless you have some resource-IDs that must be resolved to strings in your models, no need for that.

3) Grab data from the message which is internally required for the MessageController (like packet protection)
--> I don't any data from your update package is interesting to track for the message-controller

skittles1:
> This looks like the Json dll is linked from the ircbot subfolder into the core-lib?

Yeah, I'm thinking this should be in a different location but not sure where.

> I'm not a big fan of additional dependencies on the core-lib due to portability (unless it's absoluteltey needed). Not sure about the portability of Newtonsoft JSON library. 

From what I understand it works on all platforms but solid information on parsing JSON in C# and what works where (and how) is hard to come by. On http://www.newtonsoft.com/json it says "Run Anywhere" but as I don't know anything about Xamarin/Mono/MonoDroid or how to make things compatible with them, I'm not sure if adding a reference to the one dll is enough, or how to check if what I've done is compatible with other platforms. Also, I think we probably need to update the one in the repo currently.

> I have to admin I never used JSON in .net - but I really kinda expected there are some internal framework classes to deal with JSON nicely.

There are, but require more references that from what I can tell won't necessarily work on other platforms or different versions of .NET. The small amount of information available on these generally say to use JSON.NET (Newtonsoft).

cyberjunk:
I had a quick review on the 0.8.3 release of the lib here:
https://github.com/JamesNK/Newtonsoft.Json/releases

It seems they provide binaries from net 2.0 up to net 4.5 as well as "portable" library versions.
I think the preferred one to use would either the Bin/Net40 or the Bin/Portable40 from this download.

Not sure about referencing "Portable" libraries, the Meridian59 corelib itself is NOT a "portable library-project"....

I guess a good place to put it would be a subfolder "Meridian59/lib/"

skittles1:
OK added the Portable40 dll and changed the reference. Can confirm it works on Windows, at least :)

skittles1:
With closing the client, I feel like UIDownloadBar.cpp probably shouldn't be doing that (perhaps we need a download bar for something else in the future that doesn't close the client). I think UpdaterLaunch() in DownloadHandler.cs should probably do it but there doesn't seem to be a good method of accessing IsRunning from there. Should I add a method to RootClient.cs to set it to false and call it from UpdaterLaunch()?

Meridian59/Meridian59.csproj
```diff
@@ -81,7 +81,12 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Newtonsoft.Json, Version=4.5.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
```
cyberjunk: Are you sure you added the Portable40 variant?
Not sure, just saw this and it looks a bit weird "Version=4.5.0.0"

cyberjunk:
>> With closing the client, I feel like UIDownloadBar.cpp probably shouldn't be doing that (perhaps we need a download bar for something else in the future that doesn't close the client). I think UpdaterLaunch() in DownloadHandler.cs should probably do it but there doesn't seem to be a good method of accessing IsRunning from there.

Add an event to your DownloadHandler class. Name it like "ExitRequestEvent" or something. Then raise it when it is time to shutdown the client. In your BaseClient.cs change, once created the DownloadHandler object instance, register a listener to the event you created. In the listener function flip the IsRunning to false

cyberjunk:
I'm not sure how much work it would be, but I think there is an huge portion of code which could go in directly. If possible, please setup a dedicated pull which contains:

1) The added JSON library + the reference to it in the project Meridian59 / corelib
2) The new message-type enum value and the new message-class as well as the according changes in MessageController.cs
3) The MD5 function in MeridianMD5
4) ...

Then we can figure out the rest about the final update process

skittles1:
Yes, that will make it a bit more readable :) Also the Newtonsoft dll is the new one, I'm not sure why it comes up with 4.5.0.0 (I tried removing the references from the "recently used" list and readding it, but got the same result).

WIll open the new PR in a few min.

skittles1:
So this PR is now the remaining code after splitting off the other stuff.

Working on the client exiting event. Think I get it, the EventHandler is in DownloadHandler but the listener is added in BaseClient.cs?

Meridian59.Ogre.Client/OgreClient.cpp
```diff
@@ -826,10 +826,13 @@ namespace Meridian59 { namespace Ogre
...
-      // tell user about mismatching major/minor version
-      ::System::Windows::Forms::MessageBox::Show(APPVERSIONMISMATCH);
-
+      // Disconnect from server.
       Disconnect();
+      // Set UI mode to download, start download bar.
+      Data->UIMode = UIMode::Download;
+      ControllerUI::DownloadBar::Start();
+      // Download the client patch cache and updater, launch it and disconnect.
+      DownloadHandler->DownloadClientPatch(Message->ClientPatchInfo);
```
cyberjunk: I think this code-piece/Message handler should probably go into the core-library in class BaseClient.cs.
Everything is available from there and furthermore BaseClient.cs is the one which got the DownloadHandler instance.
This way the code would also be used/executed by the Bots...

Meridian59/Protocol/DownloadHandler.cs
```diff
@@ -0,0 +1,291 @@
...
+
+            // Check file length.
+            if (stream.Length != PatchFile.Length)
+            {
+                stream.Close();
+                return false;
+            }
+
+            // Number of bytes to read.
+            long numBytesToRead = 1024;
```
cyberjunk: Please define this 1024 as a constant at the top of the file and reuse the constant in the "<" check below.

PR #123, opened 2016-04-14
Add ClientPatchInfo/ClientPatchMessage data, JSON.NET lib.

skittles1:
Server sends AP_CLIENT_PATCH which is handled as ClientPatchMessage, the
data sent is stored in a ClientPatchInfo instance. Ogre is handling this by just
showing the "app version mismatch" dialog until downloading is implemented.

Added function to return the unaltered MD5 hash of an input.

Added portable Newtonsoft JSON parsing library.


Meridian59.Ogre.Client/OgreClient.cpp
```diff
@@ -824,6 +824,14 @@ namespace Meridian59 { namespace Ogre
 		Disconnect();
 	};
 
+   void OgreClient::HandleClientPatchMessage(ClientPatchMessage^ Message)
```
cyberjunk: Can you fix the formatting here. I'm really sorry for the spaces<->tabs differences between the old client and new, annoyed me a couple of times too

Meridian59/Common/MeridianMD5.cs
```diff
@@ -49,6 +49,14 @@ public static byte[] ComputeMD5(byte[] Input)
             return bytes;
         }
 
+        public static byte[] ComputeClientPatchMD5(byte[] Input)
```
cyberjunk: I think this is a generic/untouched MD5 right?
May be the name should rather be "ComputeGenericMD5()" or something, because without looking at the implementation I would have expected some special difference for the Patcher.

Meridian59/Protocol/Enums/MessageTypeLoginMode.cs
```diff
@@ -33,6 +33,7 @@ public enum MessageTypeLoginMode : byte
         GetAll          = 9,
         ReqMenu         = 10,
         AdminNote       = 11,
+        ClientPatch     = 12,
```
cyberjunk: Please add a comment here for which branch this is valid.
Like "Only valid for MeridianNext" or something

cyberjunk:
Looks fine so far, just three very minor things.. will merge right after that

Meridian59.Ogre.Client/OgreClient.cpp
skittles1: I wasn't sure which I should use http://i.imgur.com/KrOODna.png :p

Tabs?

cyberjunk: That's weird. But you're welcome to reformat the whole project to 2 spaces :-D

skittles1:
I left the spaces/tabs as it was, would be happy at a later date to format the whole file :)

PR #125, opened 2016-04-20
Convert DownloadHandler/ClientPatchInfo to use new process.

skittles1:
DownloadHandler now just downloads and runs the updater. Arguments given to updater are also saved to dlinfo.txt. Updater and dlinfo.txt saved to parent directory, and is run with this as the working directory.


skittles1:
Outdated due to https://github.com/cyberjunk/meridian59-dotnet/commit/ca74bc049e410fd0500d05802af96e937566222b

PR #132, opened 2016-05-04
Ketzer2002 switch language   issue#118

Ketzer2002:
Finished translating the static UI Resources into german language


cyberjunk:
Closing this. Very old. There is a solution by now that allows switching ui strings more dynamically.

PR #133, opened 2016-05-07
Editing gui options Step 1

Ketzer2002:


Meridian59/Common/Config.cs
```diff
@@ -231,6 +238,22 @@ public bool PreloadMusic
         }
 
         /// <summary>
+        /// Which language to load.
+        /// </summary>
+        public LanguageCode PreloadMusic
```
cyberjunk: The property name "PreloadMusic" doesn't look right here.

Meridian59.Ogre.Client/OgreClientConfig.cpp
```diff
@@ -233,6 +233,10 @@ namespace Meridian59 { namespace Ogre
 		/*  2) UI
 		/******************************************************************************/
 
+		// selectedlanguage
+		node = Document->DocumentElement->SelectSingleNode("/" + XMLTAG_CONFIGURATION + "/" + TAG_UI + "/" + TAG_SELECTEDLANGUAGE + "/" + TAG_LANGUAGE);
+		SelectedLanguage = (node && node->Attributes[XMLATTRIB_VALUE]) ? node->Attributes[XMLATTRIB_VALUE]->Value : DEFAULTVAL_INPUT_LANGUAGE;
```
cyberjunk: The language config value should be completely handled by the corelib's "Config.cs" class, because it will be needed/useful for other clients as well.

Therefore please don't put the value inside the <ui> xml structure, it should be somewhere at the top where the base-class reads its data (like somewhere right after <resources> for example)

Meridian59.Ogre.Client/configuration.template.xml
```diff
@@ -152,9 +152,16 @@
     <soundvolume value="10.0" />
     <disableloopsounds value="false" />
   </engine>
+  
   <ui>
-    <layout />
+    <selectedlanguage>
```
cyberjunk: This is what I mentioned in the other comment.
Please make sure the value is somewhere at the beginning when the base-class "Config.cs" reads its data.
Then read and write the property from the base-class. I don't think any changes in the subclass OgreClientConfig should be needed.

Resources/ui/layouts/Meridian59.layout
```diff
@@ -2481,6 +2481,19 @@
 								<Property name="Area" value="{{0,240},{0,280},{0,360},{0,300}}" />	
 							</Window>
 
+							<!-- SelectedLanguage -->
```
cyberjunk: Not 100% sure what's the parent element here.
I had a quick review on the UI window, and I guess the "Gameplay" section would probably best fit to add another combobox to pick the language?

Meridian59.Ogre.Client/OgreClientConfig.h
```diff
@@ -87,6 +88,8 @@ namespace Meridian59 { namespace Ogre
 		literal ::System::String^ TAG_SOUNDVOLUME = "soundvolume";
 		literal ::System::String^ TAG_DISABLELOOPSOUNDS = "disableloopsounds";
 		literal ::System::String^ TAG_UI = "ui";
+		literal ::System::String^ TAG_SELECTEDLANGUAGE = "selectedlanguage";
+		literal ::System::String^ TAG_LANGUAGE = "language";
```
cyberjunk: You have specified both of these constants in the base-class "Config.cs" already. No need to it in the subclass again (as mentioned I think you don't have to touch the OgreClientConfig class at all).

Remember that OgreClientConfig class in the C++ CLR OgreClient project just derives from Common.Config class in the C# corelib. The base class is the one which starts reading and writing data from XML and then hands over to the sub-class, to read and write even more data from/to xml. The Language setting reading/writing as well as its constants and properties should be completely inside the Base class.

Ketzer2002: Removed the String literals out of the derived class

Resources/ui/layouts/Meridian59.layout
Ketzer2002: will change this to the general tab

Ketzer2002: Parent Element is Engine -> Tab is general

Meridian59/Common/Config.cs
Ketzer2002: Typo deleted

Meridian59.Ogre.Client/OgreClientConfig.cpp
Ketzer2002: Changed it

Meridian59.Ogre.Client/configuration.template.xml
Ketzer2002: Done, Commit and Pull-Request follows

Ketzer2002:
An Error since I'm editing with Visual Studio 2013 Express occured, I'm still unable to test the Client in the dev environment... are ther more external dependencies in the project?

maybe I've done something wrong to load the engine... my VS says:

Fehler	58	error LNK1104: Datei "boost_d.lib" kann nicht geöffnet werden.	D:\Git\meridian59-dotnet\Meridian59.Ogre.Client\LINK	Meridian59.Ogre.Client

skittles1:
I can answer this one... you need to build the Engine solution in the \Engine\ folder, both debug/release x86/x64 as the libraries are needed for the Ogre client.

Ketzer2002:
Ok, thanks, I'll have a try...

Ketzer2002:
mhhh, I get an error, when I try to compile the Engine on X64 -> DEBUG and RELEASE

Fehler	1	error RC1015: cannot open include file 'afxres.h'.	D:\Git\meridian59-dotnet\Engine\src\ois-1.4.0\OIS.rc	10	1	OIS

I read in a forum, that the Express Edition of VS doesn't support this MFC class... maybe this is usefull for other contributors.

cyberjunk:
>> cannot open include file 'afxres.h'. D:\Git\meridian59-dotnet\Engine\src\ois-1.4.0\OIS.rc 10 1 OIS

Pretty much sounds like this issue in the past:
https://github.com/cyberjunk/meridian59-engine/commit
/a342adc3873ed3d97e6e1fd79917572dd00da1cd

PR #138, opened 2016-05-20
Statusbar Languagestrings

Ketzer2002:


cyberjunk:
Nice. BTW: You don't really need to group all the single words like "Add", "Delete", and and.
You can just put them into the main / general set (probably even better because they can be reused, unlike the tooltip descriptions etc.)

cyberjunk:
Let me know when this is ready to merge. Do you want to add more here or do them in a different pull?

PR #139, opened 2016-05-27
Hash rsb file on login, send hash to server.

skittles1:
The rsb file is hashed on loading and the hash sent to the server during login. The server will trigger a client update if the rsb hash does not match the server's loaded one.

This is the .NET implementation of https://github.com/OpenMeridian105/Meridian59/pull/172. If this approach is correct I'd probably move the file hash function to Util.cs in case we need it elsewhere.

### DIFF
--- a/Meridian59.Ogre.Client/OgreClient.h
+++ b/Meridian59.Ogre.Client/OgreClient.h
@@ -222,7 +222,7 @@ namespace Meridian59 { namespace Ogre
 
 		property unsigned char AppVersionMinor
 		{ 
-			public: virtual unsigned char get() override { return 6; } 			
+			public: virtual unsigned char get() override { return 7; } 			
 		};
 		
 		property ::Ogre::Root* Root 
--- a/Meridian59/Client/BaseClient.cs
+++ b/Meridian59/Client/BaseClient.cs
@@ -730,7 +730,7 @@ namespace Meridian59.Client
         {
             // create message instance
             LoginMessage message = new LoginMessage(
-                Username, Password,
+                Username, Password, ResourceManager.RsbHash,
                 AppVersionMajor, AppVersionMinor,
                 LoginMessage.WINTYPE_NT, 6, 1, 512000000, LoginMessage.CPUTYPE_PENTIUM,
                 MeridianExeCRCs.NEWCLIENTDETECT,
--- a/Meridian59/Common/Hash128Bit.cs
+++ b/Meridian59/Common/Hash128Bit.cs
@@ -14,12 +14,12 @@
  If not, see http://www.gnu.org/licenses/.
 */
 
-namespace Meridian59.Protocol.Structs
+namespace Meridian59.Common
 {
     /// <summary>
     /// A struct to store 128-Bit hashes in 4x 32-Bit blocks.
     /// </summary>
-    public struct PasswordHash
+    public struct Hash128Bit
     {
         public uint HASH1;
         public uint HASH2;
--- a/Meridian59/Common/MeridianMD5.cs
+++ b/Meridian59/Common/MeridianMD5.cs
@@ -15,6 +15,7 @@
 */
 
 using System;
+using System.IO;
 using System.Text;
 using System.Security.Cryptography;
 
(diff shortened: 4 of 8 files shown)


cyberjunk:
I see why you want to double check that the RSB file is up to date and valid, since it's very essential...
As you mention, please move the "CalculateFileHash(string file)" at a more generic place, like the Util and please make it return a byte[] rather than a string.

I don't really get why you transfer the MD5 as string (while the same package contains a binary MD5 as password...). This seems a bit unhandy and inconsistent to me?

cyberjunk:
One more thing:
The Login message already has the (truncated, 16bit) CRC for the meridian.exe in it.
However, I think last time I checked this was ignored on the server-side...

It might be an idea to extend your approach here towards a combined CRC for the "essential files".

So instead of creating an additional CRC for the RSB file only, create it on the EXE and the RSB and then OR the hashes together and send this (for Ogre you can OR all three, both x86/x86 exe and the RSB). This way you could also catch corrupted EXE files, not only corrupted RSB files.

skittles1:
I want to keep the RSB checking fairly simple (involving just what files the server normally has available) because server 112 will also be using this in their update process and possibly other servers will eventually too. I think separately we can modify the CRC so the server can optionally check for that for classic/Ogre.

I see what you mean with the hash, I've treated it as a hex string on the server for (manual) comparison to the hash in the patchinfo.txt file but it does overcomplicate it a bit. Would you recommend just sending the raw MD5 hash to the server?

cyberjunk:
>> I want to keep the RSB checking fairly simple (involving just what files the server normally has available) because server 112 will also be using this in their update process and possibly other servers will eventually too.

Ok then keep the CRC to the RSB only.

>> I see what you mean with the hash, I've treated it as a hex string on the server for (manual) comparison to the hash in the patchinfo.txt file but it does overcomplicate it a bit.

I see why you want to have a hex-string for comparison. But it should be possible to turn it into a hex-string in the UI later? It seems a bit weird to have two MD5 in the same message (password+rsbhash), but both encoded differently..

cyberjunk:
Oh I forgot: This also needs some VANILLA tag handling, otherwise it fill break vanilla completely I guess.

skittles1:
> This also needs some VANILLA tag handling, otherwise it fill break vanilla completely I guess.

The server stops checking the AP_LOGIN message after the password, I'll add the VANILLA tag for safety but this client was able to log in to 103 as well as a local modified server with no problems.

cyberjunk:
>> The server stops checking the AP_LOGIN message after the password, I'll add the VANILLA tag for safety but this client was able to log in to 103 as well as a local modified server with no problems.

Please add a vanilla check at least in the ByteLength and Read/Write methods. Can keep a common constructor and just ignore the parameter etc.

skittles1:
Not entirely sure how to proceed with the changes to LoginMessage, should I make a new struct like PasswordHash for RsbHash to send to the LoginMessage constructor, or keep it as byte[]? The new commit has the changes to ResourceManager and Util.

cyberjunk:
I think the PasswordHash struct should probably be renamed to something more generic, like "Hash128Bit" or something and then be moved to the Common namespace.
Then you could use this type for both MD5 (password+rsb) in the LoginMessage..

On the other hand byte[] is basically just fine, I created this PasswordHash struct back then because I didn't like the variable length on the byte[]...

skittles1:
Changed it to Hash128Bit in e303d11, does that look ok? Also do you think CalculateFileMD5() should return this struct (or an empty one) instead of byte[] for that reason? Right now if the file doesn't exist it returns `new byte[0]` which I'm thinking might not really be right, especially if we try to encode this into the struct later.

cyberjunk:
>> Changed it to Hash128Bit in e303d11, does that look ok?

Yes looks good. But I just realized there is a "ComputeGenericMD5(byte[] Input)" in the MeridianMD5.cs class already. Probably this util function should be moved there as "ComputeGenericFileMD5" for something.. ?

skittles1:
OK moved the function, but still have it returning byte[] rather than Hash128Bit. Should I have the RsbHash as Hash128Bit in ResourceManager and transfer it that way to LoginMessage, or should I keep it as byte[] (and transfer it to LoginMessage that way) and only encode Hash128Bit in the same manner as the password is handled?

cyberjunk:
>> OK moved the function, but still have it returning byte[] rather than Hash128Bit. Should I have the RsbHash as Hash128Bit in ResourceManager and transfer it that way to LoginMessage...

I think having the functions return the Hash128Bit struct instead of byte[] would probably be a bit cleaner, but I dunno what more changes it would cause at other places and it's technically not part of this pull/your job (unless you want to). It looks fine to me so far, so I can merge, until you want to add anything....

cyberjunk:
Wait. One thing about the encoding of the password. So you prefer to keeping it string encoded?

skittles1:
Have time to come to IRC today? Probably easier to discuss it that way :)

PR #177, opened 2016-08-25
A few more visual enhancements.

MorbusM59:
More unified color scheme, more rounded look and a highly readable chat background, that still works with the rest of the look. Also streamlined a few other elements (e.g. target and room enchants window now have the same dimensions as the combined player/buff window.)
Updated screenshot:
ogre20160831


MorbusM59:
This can be merged right away. Not adding anything else for now. :)

MorbusM59:
This can be merged right away. Not adding anything else for now. :)

cyberjunk:
I don't get that:
>> target and room enchants window now mirror player and buffs window 100%

In your screenshot it seems the room enchantments are now displayed in the target panel.
What if you don't have a current target? You don't see the room enchantments?

I don't really like that...

MorbusM59:
Nope, it's still its own window. I just changed the dimensions of the target window and the room enchants window to match the dimensions of the player and the buffs window (which are still fused together).

As I said before, I'd like the buffs window to be its own window as well, just like the room enchants window. I haven't got around to that though.

skittles1:
What about putting the options in the top-right corner as a gear icon or something, and having a big connect bar underneath the login options? Just an idea (I tried to click options to log in also).

cyberjunk:
>> (I tried to click options to log in also)

Hehe. I saw this happened to Ambermoon in the video. And it happened to me as well.
Actually I like the idea of splitting them up like Delerium suggests...

We already got an icon for the Options...
May be just place this somewhere?

MorbusM59:
Could absolutely do that.

The reasoning I have is, that the last field should be the one you hit when everything else is set. So first you do your option, then you confirm that everything is alright by hitting connect. Consequently, I'd also have options as the bottom of two buttons. I guess most people feel differently though.

It's probably the same as with clapping. I clap with my right hand into my left hand, which is below. Most people clap into their right hand. :P

MorbusM59:
options1
options2

Changed options to a little wrench symbol in the login screen and decided to go with the same icon in the game. To minimize confusion with the rather similar activate icon, I changed that one to look a bit less like an options icon. See above.

MorbusM59:
Think I spotted another little issue that popped up after "fixing" the warning... Text in the target window is no longer centered with line break. Working on it.

MorbusM59:
And with small icons...
issues2

MorbusM59:
Alright, done now. Ready to merge.

MorbusM59:
Seems like a pretty solid experience from the start now. I can remove the last commit if that is desired. I'd highly recommend it though. Feedback on the current default controls has been overwhelmingly negative so far.

I can understand the desire to have turn left and right keys. We should really change those turn keys to strafe keys while RMB is held down though. As long as that is not the case, the default mapping for A and D must be stafe left/right.

skittles1:
BTW when this pull is done I'd like to release the updated client to users as Ogre client 1.0.4.5 if you both think it's ready.

MorbusM59:
I'd like to think that I'm done with this one. :)

skittles1:
Testing out the changes, found some issues on release x86 build (haven't tried x64 yet):
<p align="center">
   
</p>
The room flag icon doesn't draw inside the bar, the options menu draws on top of chat but underneath the action buttons (prob should be on top of everything), and the inventory doesn't draw fully when I log in (scrolling fixes it).

MorbusM59:
Also removed any AlwaysOnTop flags in layout since they didn't help whatsoever and made the UI behave in unexpected ways.

cyberjunk:
Are you sure that one row for personal enchantments is enough?
If you add some more exotic buffs and possibly some debuffs in PvP, I'm not so sure...

Please at least test the behaviour for the case when there are more buffs active than can be displayed.
I'm just afraid of some code failing here...

Meridian59.Ogre.Client/OgreClientConfig.h
```diff
@@ -103,6 +118,21 @@ namespace Meridian59 { namespace Ogre
 		literal ::System::String^ TAG_ACTIONBUTTONGRID = "actionbuttongrid";
 		literal ::System::String^ TAG_ONLINEPLAYERS = "onlineplayers";
 		literal ::System::String^ TAG_ROOMOBJECTS = "roomobjects";
+		literal ::System::String^ TAG_AVATARVISIBILITY = "avatarvisibility";
+		literal ::System::String^ TAG_TARGETVISIBILITY = "avatarvisibility";
```
cyberjunk: This is incorrect :)
Likely they overwrite each other

MorbusM59:
Geez thanks! That's why target was behaving weirdly. I set it to default off because of that. (And also because it makes fairly little sense at the beginning.

Anyway, gonna check behavior. Thanks a bunch!

MorbusM59:
Everything should be fixed now.

MorbusM59:
Is there anything else that needs changing? If so, please say so. This is the last week of my vacation and I'll have very little time to work on things in the next few months.

cyberjunk:
>> This is the last week of my vacation and I'll have very little time to work on things in the next few months.

Welcome to my current life hehe :-(
I wanted to try out the changes, but I didn't really had time to so far.

I think it looks all good, except I don't feel very comfortable with the new default key bindings.

MorbusM59:
Rebased stuff and squashed/fixuped some commits. Also removed the keybinding commit. Not sure why your commit is in there now. IQ too low to use Git I'm afraid. Sorry.

cyberjunk:
My room enchantments seem misplaced (old config..)

MorbusM59:
Yeah, the old config isn't working and while you have the old config file up, it will use that value. Not sure what to do about that... I'm still not sure how the config file is handled. If I delete mine and start the game, it crashes. If I remove elements, such as the layout part, the game rewrites it.

cyberjunk:
>> If I delete mine and start the game, it crashes. 

Wow. That should definitely not happen. The game is supposed to startup without the configuration.xml file using default values.

MorbusM59:
Could you maybe log on to irc for 10 mins to sort that out?

Issue #150, opened 2016-07-16
Persistent window configuration upon logging.


MorbusM59:
It would be nice if the exact layout was persistent through logins. That includes, position, the information about whether a window is open or not and it's stack order (mostly chat appearing in front of the weapon overlay etc.) Maybe we should simply send the weapon and shield overlay to the background though.

cyberjunk:
The Open/Close state is easy, just needs to be saved/loaded with the position/size values.
For Z-Ordering I can't say for sure right now.

cyberjunk:
The open/closed state is easy.

1) Add boolean vars for visiblity here (for each of the Rects listed there)
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/OgreClientConfig.h#L222

2) Save visiblity state from UI objects to new booleans vars created in (1):
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/ControllerUI.cpp#L511

3) Apply visibility state by modifying this block (add second && configvar cond. to == playing)
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/ControllerUI.cpp#L583

4) Persistenly write config values to XML by modifying/adding here:
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/OgreClientConfig.cpp#L673

5) Read boolean visibility values from XML by modifying here:
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/OgreClientConfig.cpp#L242

MorbusM59:
Solved.

PR #181, opened 2016-08-31
Setting default keybindings to be in line with what people expect.

MorbusM59:
At least those that are familiar with other RPGs. That means that A and D are mapped as strafing left and right while rotating left and right has been sent to the left arrow and right keys. Opening doors is assigned to Space, since I haven't met a single person who had any sort of tolerance for it being default mapped to N. Sorry Shak :P 

This also includes a provisionary, but useful standard layout that should ring familiar with Meridian players. The hotkeys 1 to 0 and F1 to F12 are mapped as well as the hotkeys Q,E,R,F,T,G,B,N,Y,X,C,V as well as the ^key.

Obviously, there is a bit of an issue with different keyboard layouts. A few keys are based on the German keyboard layout, namely y for z and ^ for ~. I decided that the potential harm of having those keys bound anyway is fairly negligible and is outweighed by the benefit.

Additionaly, some windows have been made visible by default since they tend to be used frequently. (inventory, attributes, entities, spells and actions)


skittles1:
I think having the default keybinding mirror classic is especially important for new players - the signs in Raza are based on the classic controls and when broadcasting for help with keys (mostly to open doors) the responses generally assume classic keybinding.

cyberjunk:
To change the default binding, have a look here:
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59.Ogre.Client/OISKeyBinding.cpp#L27

The configuration xml is basically only for visualiation purposes. It's not used anymore.
I haven't gone through all changes in detail, but there might be things I don't like :-p

cyberjunk:
Oh, btw. please don't just make these windows all visible by default.
We discussed a nicer approach...

I added some notes how to save them with the configuration here:
https://github.com/cyberjunk/meridian59-dotnet/issues/150

MorbusM59:
Ah yeah, right, sorry. I think my brain went poof at some point while trying to figure out how to get icons to display high resolution in the buff bar. :P Closing this crap pull then.

Issue #163, opened 2016-08-05
Some people find the right click behaviour of the client unintuitive.


MorbusM59:
It would be great if a right click always brought up the info screen of the item the right click is targeted at, even if it isn't a list item, but an object in the game world. (Right now, you have to target an object, then inspect.)

cyberjunk:
You can assign the right click to any action you want...
But I agree that the "Inspect" would likely be a better default than "Attack".

MorbusM59:
I have it set to inspect, however, you still have to target (left click) an object in the world, then right click, to bring up the info box. Ideally, the right click would both select and inspect an object. Actually, I don't see much of an issue with right click also selecting things before doing the action it is assgined to in general.

Maybe as an example: If I have a zombie targeted, then right click on a skeleton, the game will bring up the info for the zombie, not the skeleton, since the right click only triggers the inspect, not the select command.

cyberjunk:
>> however, you still have to target (left click) an object in the world, then right click, to bring up the info box.

This sounds more like a bug. The right click is supposed to be able to execute an action on sometihng which is NOT your current target (e.g. without switching it).

MorbusM59:
Interesting. I don't believe that has ever been the case in Ogre though.

MorbusM59:
Working on this at the moment.

MorbusM59:
Solved in #193 .

PR #187, opened 2016-09-09
Modified default keybindings to match classical client more.

MorbusM59:
Main differences:
Q and E no longer strafe the chracter, but are freed up as hotkeys.
A and D no longer turn the character but strafe left and right.
Turn left and right have been moved to Left Arrow and Right Arrow.
n no longer opends doors, but is a regular hotkey.
Space now opens doors.

The right click default action is now hot key 48 and is unassigend (used to be space).

That leaves us with the hotkeys Q, E, R, F, T, G, Y, X, C, V, B, N, Grave (Zirkumflex), 1 - 0  and F1-F12


MorbusM59:
Before we implement this, we should also reassign the default actions on the actionbars. So far, I've been unable to find out how to set the defaults though.

cyberjunk:
Default button configuration is here:
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Data/Lists/ActionButtonList.cs#L102

MorbusM59:
Pretty much done here. Would be nice to default assign a few spells as well (such as phase and conveyance). No idea how to do that, unfortunately.

cyberjunk:
>> Would be nice to default assign a few spells as well (such as phase and conveyance). No idea how to do that, unfortunately.

You gotta do someting like:
>> buttons.Add(new ActionButtonConfig(19, ActionButtonType.Spell, Common.Constants.ResourceStrings.Spells.Shalille.DAZZLE));

Where DAZZLE is a constant from:https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Common/Constants/ResourceStrings.cs

There are only a few spells listed so far, so you gotta add there first what you need.

MorbusM59:
Sweet thanks!

MorbusM59:
All done now! Ready to merge!

MorbusM59:
We should probably come to a decision here and either merge it or close it without merging, so that we can finally roll out the update. Of course, it would be even better if we could come to a joint solution.

Maybe both of you ( @cyberjunk and @skittles1 ) could give me a quick write up of your preferred layout (I mean including at least phase and conveyance, but ideally, meditate and blink as well). Then we can see where we have common ground and come up with a version that keeps stuff, that we don't agree on, the same, but that has a few positive changes (mostly getting spells on the default action bar).

Does that sound reasonable?

skittles1:
If I understand the code right, currently you have 'attack' as unassigned (KC_UNASSIGNED). Should attack be preassigned by default to E as in the classic client? If there's a better key for attack perhaps we should have it that way in both clients just to simplify switching between them.

MorbusM59:
Attack is assigned to the E key (action button 2). But the default right click action is no longer 13, but 48. That means that right click is no longer attack, but inspect.

cyberjunk:
>> Another issue: Our right click doesn't trigger on the hovertarget, but the actual target. Currently, you will be attacking your target, when all you want to do is face a different direction.

That is not true... The right click will be executed on your current hover target, not your current target. That's the whole idea behind it, so that you can e.g. blind someone without switching targets.

cyberjunk:
For the rest. I think it probably makes sense that classic and Ogre start with as much equal keybinding as possible, but honestly *I* would not play with this binding here :)

MorbusM59:
Alright, just tested and you are right. For the attack action it actually works like that. I've only ever used right click to inspect and as we already found, it doesn't work there. I think fixing that bug might be one of the highest priority targets we have.

As for bindings: I wouldn't really play with that exact layout either. (I wouldn't use Q, R and F for activating, resting and meditate for example). But the goal here is to create a layout that is familiar to most people and that encourages them to experiment with their own layout. (That's why I feel it's super important to have the spells on the default layout.)

Unfortunately, a lot of people don't change the default layout, so as a second goal is to provide those players with a layout that allows them to play the game without issues, even though it may not be perfectly tailored to their demands.

But yeah, the first thing I do in every game is to remap controls. I don't even try playing the default layout usually.

cyberjunk:
>> I've only ever used right click to inspect and as we already found #163 , it doesn't work there. I think fixing that bug might be one of the highest priority targets we have.

Probably not very difficult....
You gotta modify [here](https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Client/BaseClient.cs#L1532), according to [here](https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Client/BaseClient.cs#L1362)

So check for _Data.IsNextAttackApplyCastOnHighlightedObject_ and if so use the _Data.RoomObjects.HighlightedID_ instead of _Data.TargetID_

MorbusM59:
Got to work on it already. :) Inspect works, activate requires a bit more work because of the container exception.

MorbusM59:
Done. Pullrequest is up.

PR #193, opened 2016-09-13
Fixed right clicks

MorbusM59:
Inspect and activate now both work like attack, targeting highlights, then regular target and, in case neither is available, the closest inspectable/activatable object.


MorbusM59:
PS: Also fixes the bug of failed access spam if a container is out of range.

cyberjunk:
Nice, but:

1)
>> ClosestInspectableInFront()
ClosestActivatableInFront()

These are huge copy'n'pastes :-(
And they don't even check for the flag they claim.
This can be done better..

2)
>> SendReqActivate()

I don't really like this change, because now this method may call _SendSendObjectContents(obj.ID)_, which breaks with the function name, which directly maps to the actual network message being sent :(

cyberjunk:
Please see this commit:
https://github.com/cyberjunk/meridian59-dotnet/commit/2d2d02a95c0c8062e895d48e05170909cd5a1df7

Please remove the two functions _ClosestInspectableInFront()_ and _ClosestActivatableInFront()_ and call the new _GetClosestObjectInFront()_ with an object flags parameter set to the desired filter (e.g. Inspect and Activate, like done [here](https://github.com/cyberjunk/meridian59-dotnet/commit/2d2d02a95c0c8062e895d48e05170909cd5a1df7#diff-45b47bc918e9eac47469c6542bce4575R1383) for attackable.

MorbusM59:
Will do.

It did check for the appropriate flags though. :P Closest activatable checked for activatable or container, just inspectable checked for nothing since everything can be inspected.

As for the copy paste: For the sake of elegance, we could probably replace all the individual send requests with SendReqAction and pass on the individual actions as parameters.

Regarding SendReqActivate() calling SendSendObjectContents(obj.ID): If I don't do it that way, I have to get the three cases (target, highlight, closest) into SendSendObjectContents(), resulting in checking the same thing twice. If I don't, SendSendObjectContents() will send the contents of the current target, and never the highlight or the closest object. Should I do it that way?

cyberjunk:
I've added the implementation for _ReqLook_ [here](https://github.com/cyberjunk/meridian59-dotnet/commit/8525441c4bf7d962d91b1020c5cd263339f71d48)

There is a couple of more places this could be useful.
I have not added it for _Activate_ yet, also could be useful for _Buy_

MorbusM59:
I've been working on a bit of a generalization as well. Particularly, I've written a message that provides the selected target to the sendrequest, be it highlight, actual target or closest in front, so that we don't have that check separately in every send request. There are a few different cases to look at, especially activate doubling for activation and opening containers. I think I'll have a decent solution out over the weekend though.

MorbusM59:
Obsolete.

PR #195, opened 2016-09-16
Improvements to most actions.

MorbusM59:
- All actions that interact with a target now also accept highlight
targets and consider objects in proximity if neither highlight nor
target are available. This includes, attack, inspect, activate, buy and
trade.
- Streamlined code a bit by outsourcing target selection to
GetSelectedObject(...)
- Fixed a missing tick check for SendSendObjectContents(ID)
- Made looting more user-friendly: If a lootable target (not a corpse!) is selected,
loots the target, else, brings up the loot window. Pushing the button
again closes the loot window again. Holding down the mod key still works
as loot all.

### DIFF
--- a/Meridian59/Client/BaseClient.cs
+++ b/Meridian59/Client/BaseClient.cs
@@ -1355,39 +1355,23 @@ namespace Meridian59.Client
                 GameTick.DidReqAction();
             }
         }
-      
+
         /// <summary>
-        /// Requests to hit your current target or highlighted target
+        /// Requests to hit your current target, highlighted target or closest
+        /// attackable object.
         /// </summary>
         public virtual void SendReqAttackMessage()
         {
-            // attack highlighted
-            if (Data.IsNextAttackApplyCastOnHighlightedObject)
-            {
-                if (ObjectID.IsValid(Data.RoomObjects.HighlightedID))               
-                    SendReqAttackMessage(Data.RoomObjects.HighlightedID);
-                
-                // unset nexttarget on highlight
-                Data.IsNextAttackApplyCastOnHighlightedObject = false;
-            }
-           
-            // attack current target
-            else if (Data.TargetObject != null)
-            {
-                SendReqAttackMessage(Data.TargetObject.ID);
-            }     
-      
-            // else attack attackable in front
-            else
-            {
-                ObjectFlags attackFlags  = new ObjectFlags();
-                attackFlags.IsAttackable = true;
+            // Set the appropriate flags to check for.
+            ObjectFlags FilterFlag = new ObjectFlags();
+            FilterFlag.IsAttackable = true;
+        
+            // Let's figure out our target.
+            RoomObject SelectedObject = GetSelectedObject(FilterFlag);
 
-                RoomObject obj = Data.GetClosestObjectInFront(attackFlags);
-
-                if (obj != null)
-                    SendReqAttackMessage(obj.ID);
-            }
+            // If we found something, interact with it.
+            if (SelectedObject != null)
+                SendReqAttackMessage(SelectedObject.ID);
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


MorbusM59:
I hope this pull request makes a bit more sense. I tried to keep all the previous loot functionality while eliminating the annoying parts.

I can't guarantee that certain parts of code couldn't be more elegant and brief, but I tried my best with the 5 hours of C# experience I've had so far. :P

Meridian59/Client/BaseClient.cs
```diff
@@ -1694,7 +1660,72 @@ public virtual void SendReqCastMessage(SpellObject Spell)
...
+
+                    // No filter provided, return highlight and reset IsNext...
+                    if (FilterFlag == null)
+                    {
+                        Data.IsNextAttackApplyCastOnHighlightedObject = false;
+                        return SelectedObject;
+                    }
+
+                    // If highlight matches our flag, return it and reset IsNext...
+                    if (FilterFlag.IsSubset(SelectedObject.Flags))
```
cyberjunk: This can throw a NULL exception, because _SelectedObject_ might be NULL.

Meridian59/Client/BaseClient.cs
```diff
@@ -1694,7 +1660,72 @@ public virtual void SendReqCastMessage(SpellObject Spell)
...
+                    // null.
+                    else
+                        return null;
+                }
+            }
+
+            // If we don't have a highlight, but a regular target, pick that one instead.
+            else if (Data.TargetObject != null)
+            {
+                SelectedObject = (RoomObject) Data.TargetObject;
```
cyberjunk: The TargetObject is of type _ObjectBase_, you can't just cast to _RoomObject_.
For example if you target an item from your inventory, it would be of type _InventoryObject_...

cyberjunk:
It seems reasonable to have a reusable function doing the selection (highlight/target/closest). So I added something like your _GetSelectedObject_ here (however I implemented it in DataController class, seems a better fit):

https://github.com/cyberjunk/meridian59-dotnet/commit/30e1f2766c574179068ba03ec80dc1754329bed9

cyberjunk:
Then [this](https://github.com/cyberjunk/meridian59-dotnet/commit/8007b06311a31d6bced355356ddf30d5b3aa537c) replaces the duplicated code in _SendReqAttackMessage()_ and _SendReqLookMessage()_ with the new function

MorbusM59:
Alright, pulled your latest master and created a new branch. Will try to carry over the changes I made to the new branch while fixing the issues you pointed out.

cyberjunk:
Wait a moment. I'll add another commit soon...

MorbusM59:
Alrighty. :)

I'd kinda have to modify your data controller method to include the PassOn functionality. Else, activate's double function won't work. :/

cyberjunk:
Activate is a bit nasty. That's due to the fact that it is two different flags and technically should be two different actions. But at some point I decided to merge them both into the Activate action, now causing trouble or at least ugly code.

MorbusM59:
Yeah, as you've seen, I ran into the issue of losing the highlight action flag on the first action. :/
It is kinda convenient for the user though. Ideally, I'd say that opening doors, containers and activating could all be the same "interact" action.

MorbusM59:
Oh and a little question about the null eception.
if (ObjectID.IsValid(Data.RoomObjects.HighlightedID))
{
    SelectedObject = Data.RoomObjects.GetHighlightedItem();
}

Doesn't that guarantee that SelectedObject != null?

cyberjunk:
Ok, this is the commit allowing buy, activate etc. on the closest/highlighted object:
https://github.com/cyberjunk/meridian59-dotnet/commit/b11c00ed7bdffc682aada1988c9589b4d859e594

It seems I was fine without any 'PassOn'...

cyberjunk:
I've also added the:
>> Fixed a missing tick check for SendSendObjectContents(ID)

However I have not integrated the additional _Loot_ behaviour.

cyberjunk:
I think a generalization of the loot adjustements you proposed is that sometimes you want to make sure that the current target or highlighted object matches the required flags, or otherwise rather fall back to the closest object case.

I think the best solution is to improve 'GetInteractObject()' in this case.

MorbusM59:
Regarding PassOn: The issue arises when you try to right click a lever or flagpole. Your initial click will trigger the first attempt to get a highlight target. However, since the target isn't compatible with a container, it will pass on to ReqSendActivate(). However, ReqSendActivate() doesn't have the information that you wanted to interact with your highlight object, so it will simply try to activate your regular target or the closest compatible target in front.

As for looting: I believe that too much of a smart loot system will be more confusing than helpful. You generally don't want to be looting anything you haven't approved of looting, unless you went for a "loot all". Maybe I misunderstood you, but I don't think there is any use for a "loot closest object", since, for the player, that is mostly the same as "loot a random object".

I played a bit with my current loot settings, and to me, it keeps full control while also making it very easy to handle looting just with one button (and the modifier key). I'm not sure further modification is necessary really.

I'll just see what you did and work from there. :)

cyberjunk:
>> Regarding PassOn: The issue arises when you try to right click a lever or flagpole. Your initial click will trigger the first attempt to get a highlight target. However, since the target isn't compatible with a container,...

I guess you mean the case handler of _AvatarAction.Activate_: I think you missed the possibility to provide several flags which are 'OR' in the checks. This way it returns either the closest activatable object or container object. Then based on flags it decides which sender to call (see commits..)

MorbusM59:
Oh, nvm then. :) Are you all done btw?

cyberjunk:
>> Oh, nvm then. :) Are you all done btw?

No, one more change for 'GetInteractObject()' in progress :)

MorbusM59:
Alrighty! Thanks for all the work!

MorbusM59:
Nathak is aking for you in IRC btw.

cyberjunk:
Ok I'm done now.
Have a look at the latest commit here:
https://github.com/cyberjunk/meridian59-dotnet/commit/c71c9f06e14fbbf86f313dd1a6e833c006aa7519

I improved the _GetInteractObject()_ so that you can enable flags validation on the current target also and if fails return null or fall back to the closest filter match.

By default I've enabled this behaviour for BUY, TRADE and ACTIVATE.
So if you are standing in front of Frisconar and you accidentially have a current target, which you can't buy from (e.g. inventory, decoration) and you press or click BUY or TRADE, it will still initialize the dialog with Frisconar.

That's a bit tricky from the user interface consistency, because we now trigger an action on something different silently, while the user has an explicit target, but I think this is way more convenient.

MorbusM59:
Obsolete.

PR #204, opened 2016-12-05
Add 'balance' and 'appeal' chat commands, fix secret guild creation.

skittles1:
Added missing 'balance' and 'appeal' chat commands/user commands.

Before this gets merged, I was wondering if I should add the 'time' chat command that gets a printout of the time from the server. I know Ogre has its own game time in the titlebar and this should be in sync outside modified server time (i.e. testing, live servers are in sync) so this would mainly be for completeness.

Also fixed secret guild creation (wasn't working due to typo).

### DIFF
--- a/Meridian59/Client/BaseClient.cs
+++ b/Meridian59/Client/BaseClient.cs
@@ -931,6 +931,19 @@ namespace Meridian59.Client
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


cyberjunk:
Nice, thanks for looking into this.

PR #205, opened 2016-12-06
Fix icky cave exit, slope in outside bow hut map.

### LINKED ISSUE #184: Transition from Spooky etc. Cave to Twisted Woods is still very glitchy.
Most people take 3 to 5 tries to actually transition from one zone to the other.

skittles1:
Icky cave is difficult to exit in ogre client due to the small space
between the southern bounding box of the screen and the wall, and the
walls being passthrough. Made this bigger and set the walls to
non-passthrough to stop ogre bouncing out of the invalid sector behind
the wall. Fixes #184.

In the outside bow hut map, it's possible to get stuck behind a ledge on
the left-hand side of the screen. Raised the height of the ledge so
players can walk out of it.


cyberjunk:
Thanks

PR #206, opened 2016-12-06
Fix magic items not showing up with orange names.

### LINKED ISSUE #154: Magic item coloring repoted broken
A player reported that magic items are not correctly highlighted in several UI parts anymore.

skittles1:
GetColorFor needs to check if the item has the magic item flag and if
so, color the name orange so magic weapons show up in lists and the item
description.

Fixes #154.

### DIFF
--- a/Meridian59/Drawing2D/NameColors.cs
+++ b/Meridian59/Drawing2D/NameColors.cs
@@ -47,9 +47,13 @@ namespace Meridian59.Drawing2D
         public static uint GetColorFor(ObjectFlags Flags)
         {
 #if !VANILLA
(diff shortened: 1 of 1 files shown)


cyberjunk:
Thanks! I've slightly adjusted your patch here:
https://github.com/cyberjunk/meridian59-dotnet/commit/7c9e851ac629cd2bfd3f339b5f7a5f2b4a6a81db

I kinda preferred to keep the color-picking (magic or default) separated from the opacity-correction afterwards.

skittles1:
That is nicer, thanks :)

PR #214, opened 2017-02-24
Update 2.0 changes.

skittles1:
Added updated rooms and other resources for the 2.0 update.

Changed the splash screen to match the new one, with black background and dark red subtitle text.


cyberjunk:
Very nice!
All of the PNGs are for existing BGF files, yes? Because I see no BGF files in the commits.
In case there are fully new textures, OgreClient would need both (new BGF+ new PNG) ...

skittles1:
The other bgf roomtextures aren't in the /Resources/bgftextures folder so I didn't add these ones but they are sent out in the client download. They are new bgfs from this update though.

cyberjunk:
Yes. You are right, no BGF in the repo at all.
Just put them in the patcher.

Gonna have a look at this soon :-)

Issue #147, opened 2016-06-23
Add guildhall purchase logic/ui


cyberjunk:
It's impossible to currently by a guildhall.
The message / core implementations as well as the UI are missing.

skittles1:
Issue fixed in #217 :)

PR #217, opened 2017-03-14
Add guildhall purchase UI.

skittles1:
Add Ogre user interface for buying guild halls from Frular.

Displays a message above the Buy/Cancel buttons if no guild password entered.

I wasn't entirely sure how to handle the halls in DataController, right now I have it as a list of the guild halls and use ListChanged to determine whether to show the UI or not.

Addresses issue #147


cyberjunk:
> I wasn't entirely sure how to handle the halls in DataController, right now I have it as a list of the guild halls and use ListChanged to determine whether to show the UI or not.

I usually use a _IsVisible_ property on the c# data model to be able to define visibility also from the core library. The CEGUI/OgreClient observes this property. See _GuildInfo_ or _BuyInfo_ data models for examples which have IsVisible_ and their according UI elements.

To flip the visibility I use the message-processing as trigger.
Like in _HandleBuyList(BuyListMessage Message)_ in DataController which sets visible to true.

In CEGUI you just listen for the change in the IsVisible property to actually set the CEGUI view elements visible or not...

skittles1:
Updated the original commit with the GuildHallsInfo object handling visibility and the GuildHalls list.

cyberjunk:
Seems good. Ready for merge?

skittles1:
Yes, should be ok to merge.

Issue #93, opened 2016-01-05
Add confirmation window for guild commands such as renounce and exile.


skittles1:
Leaving guild accidentally comes with a long rejoin timer so players should be asked to confirm the action before sending the renounce/exile command to the server. Opening issue before I forget about it, might work on this in a couple days if I have time (and if you haven't done it yet).

skittles1:
Changes live :)

PR #218, opened 2017-03-16
Add missing Guild UI features, confirm dialog boxes for guild UI actions.

skittles1:
#### Commit 1
Guild UI was missing an 'Abandon guildhall' button to drop the guild's hall. Added this, and also added text which is displayed instead of the guildhall options if the user has no hall.

If the user shouldn't have the tabs for GuildMaster and Shield, remove them from being displayed. Replace them if the user then tries to access the Guild menu as a guildmaster. In CEGUI there's no actual way to do this save for removing and replacing the tab windows from the TabControl (as per http://cegui.org.uk/forum/viewtopic.php?t=3736 and http://cegui.org.uk/forum/viewtopic.php?f=10&p=35580).

#### Commit 2
Make exile, renounce, disband and abandon hall all require an additional confirmation before carrying out the action. Confirm dialog box prevents interacting with other UI elements until the box is addressed.

One thing I wasn't sure of here - I use a variable (ObjectID) to keep track of which ID needs to be exiled or abdicated to since the confirm dialogs no longer have that selection data (and this seems the most elegant way of doing it).

Addresses issue #93

### DIFF
--- a/Meridian59.Ogre.Client/Constants.h
+++ b/Meridian59.Ogre.Client/Constants.h
@@ -405,6 +405,8 @@
 #define UI_NAME_GUILD_PASSWORDDESC					"Guild.PasswordDescription"
 #define UI_NAME_GUILD_PASSWORDVAL					"Guild.PasswordValue"
 #define UI_NAME_GUILD_SETPASSWORD					"Guild.SetPassword"
+#define UI_NAME_GUILD_ABANDONHALL					"Guild.AbandonHall"
+#define UI_NAME_GUILD_NOGUILDHALL					"Guild.NoGuildHall"
 #define UI_NAME_GUILD_RENOUNCE						"Guild.Renounce"
 #define UI_NAME_GUILD_SHIELDIMAGE					"Guild.ShieldImage"
 #define UI_NAME_GUILD_SHIELDCOLOR1DESC				"Guild.ShieldColor1Description"
--- a/Meridian59.Ogre.Client/ControllerUI.h
+++ b/Meridian59.Ogre.Client/ControllerUI.h
@@ -888,6 +888,8 @@ namespace Meridian59 { namespace Ogre
 			static ::CEGUI::Window* PasswordDesc = nullptr;
 			static ::CEGUI::Editbox* PasswordVal = nullptr;
 			static ::CEGUI::PushButton* SetPassword = nullptr;
+			static ::CEGUI::PushButton* AbandonHall = nullptr;
+			static ::CEGUI::Window* NoGuildHall = nullptr;
 			static ::CEGUI::Window* ShieldImage = nullptr;
 			static ::CEGUI::Window* ShieldColor1Desc = nullptr;
 			static ::CEGUI::Slider* ShieldColor1 = nullptr;
@@ -1744,6 +1746,7 @@ namespace Meridian59 { namespace Ogre
 			static bool OnRankSelectionChanged(const CEGUI::EventArgs& e);
 			static bool OnDiploSelectionChanged(const CEGUI::EventArgs& e);
 			static bool OnSetPasswordClicked(const CEGUI::EventArgs& e);
+			static bool OnAbandonHallClicked(const CEGUI::EventArgs& e);
 			static bool OnRenounceClicked(const CEGUI::EventArgs& e);
 			static bool OnExileClicked(const CEGUI::EventArgs& e);
 			static bool OnGuildShieldSettingChanged(const CEGUI::EventArgs& e);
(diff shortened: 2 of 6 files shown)


cyberjunk:
Thanks for your work on this.
Have you seen the "UIConfirmPopup.cpp" file?

It is currently used for the 'Suicide' confirmation only, but as the name suggests, I created it with the intention to turn it into a generic Yes/No confirmation prompt at some point that can trigger two defined actions fors yes/no.

There is some work to do on this to make it dynamically enough to serve the mentioned purpose, but I think there are probably several cases also outside of the guid-window which would benefit from a generic solution to provide yes/no confirmation.

cyberjunk:
This is basically how to turn the _ConfirmPopup_ defined in _ControllerUI.h_ and implemented in _UIConfirmPopup.cpp_ into a more generic solution that let's you define appropriate actions for yes and no from the caller location and therefore can be used at many places:

1) Add a public method to _ConfirmPopup_, like "Show(string text)", put in the code which is currenly in _OgreClient::Suicide()_ in _OgreClient.cpp_ to show the window and give it focus. Additionally use the text param to set a more detailed warning/message on the window within this Show(..).

2) Add two .net events to the _ref class ConfirmPopup_ in _ControllerUI.h_ I would name them like _Confirmed_ and _Cancelled_

3) In the current CEGUI handlers for the button clicks, raise the newly defined events and removed the fixed handling for suicide that is currently in there.

4) In _OgreClient::Suicide()_, assign event handlers to the newly added events that contain the code the buttons should actually execute, in this case the old suicide handling from the yes-click handler of the ConfirmPopup.

5) Make sure to remove all attached event handlers again, when the ConfirmPopup is closed.

5) Use the same technique at other places, e.g. in guildwindow, call Show(text) on the ConfirmPopup and assign handlers to the button-click for the situation the window was brought up.

skittles1:
I'll try get the generic ConfirmPopup changes done on Saturday and submit it as a separate PR if you're not already working on it.

cyberjunk:
Please have a look here for the generalization of the existing confirmation popup:
https://github.com/cyberjunk/meridian59-dotnet/commit/70597f60d4adf8c0c14ef6726491fb431e6d34f9

It should be possible to use this for any case now, just like done for suicide.
Code that should run for buttons can now be set together with showing the window.
Can be improved (e.g. make title also a param of Show(), allow selecting between yes/no and OK only buttons, ...)

You can definitely apply your layout-style, e.g. making it modal blocking the UI etc.

skittles1:
Rewrote the confirm box events for the new ConfirmPopup UI, but #219 should be merged first so this one can use the new function for the yes/no box.

skittles1:
Ok this one should be done if you want to review

cyberjunk:
I think we can find better solution than saving the ObjectID temporarily on the Guild UI class.

There are probably more cases which require additional params in the Confirm/Cancel handlers, specifically some sort of ID.

First, I think I would like the ObjectID to be moved to the ConfirmationPopup.
It seems to make more sense to save it in this context?

Second, it could also be handed over along with the Confirmed and Cancelled events, which means using some custom subclass of _EventArgs_ for these events and settings the id when raising them. Then the handlers can directly access the id ...

skittles1:
So would we use https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Common/Events/ObjectIDEvent.cs for this?

cyberjunk:
Yes, or the IntegerEventArgs:
https://github.com/cyberjunk/meridian59-dotnet/blob/master/Meridian59/Common/Events/IntegerEvent.cs

There might be cases when it's not really a blakserv ID but some other kind of ID...
You can start by moving the "ObjectID" member from Guild to ConfirmPopup first.

cyberjunk:
Ok good.
I think the ID should be reset at only one place, when the event handlers are also deleted.

Set to 0 here in ConfirmationPopup.cpp (both for confirmed and cancelled):
>> // remove handler(s)
_confirmed = nullptr;
**id = 0;**

Probably the ID should be set together with the other data in Show() methods, like an additional, possibly (last) optional argument.

skittles1:
I tried optional arguments but it didn't want to allow those on 'functions of a managed type' so I added overloads for ShowChoice and ShowOK.

cyberjunk:
>> // remove handler(s)
_confirmed = nullptr;
ID = 0;

Do this in both _RaiseConfirm() and _RaiseCancel().

Also I forgot to unset the other event.
The _RaiseConfirm() and _RaiseCancel() handlers should both have:

>>// remove handler(s)
_confirmed = nullptr;
_cancelled = nullptr;
ID = 0;

cyberjunk:
>> I tried optional arguments but it didn't want to allow those on 'functions of a managed type' so I added overloads for ShowChoice and ShowOK.

That's OK. But I think you can then just all the variant with the UINT from the variant without it and use 0 in there. Probably better than having duplicated code.

skittles1:
Was there anything else I should do for this? I had a look at changing the event type to pass the ID to the 'confirm' functions but I'm not sure if it's possible without changing all uses of the confirm/cancel handlers to IntegerEventArgs (my knowledge on events is still a bit lacking).

cyberjunk:
>> Was there anything else I should do for this?

No, it seems fine. I can do the change with attaching the ID to the event later.

Issue #48, opened 2015-05-27
Login Screen: Wrong username/PW creates a standard alert box


novemr:
You gonna lose your focus on the game window + can't see any mouse cursor, so it's hard to hit "OK" (well, if you don't use ENTER)

The Error Notice should be ingame, not just a plain alert box

![image](https://cloud.githubusercontent.com/assets/12058341/7844399/6d30c6a0-04b0-11e5-8638-33955e83c6e0.png)

cyberjunk:
Fixed in:
https://github.com/cyberjunk/meridian59-dotnet/pull/220

PR #220, opened 2017-03-19
Add proper confirm dialogs for login errors.

skittles1:
Replace the windows forms error boxes with the always-on-top confirm popup dialog (with single OK button).

Addresses issue #48


cyberjunk:
Is the single OK button triggering the Cancel event?

It's just that you named the handler:
> OnLoginErrorConfirmed(..)

But registered it for the Cancel events.

skittles1:
I should probably name it OnLoginErrorDialogClosed to reflect more closely what it does.

The OK button uses the same handlers as the No/Cancel button event.

cyberjunk:
>> The OK button uses the same handlers as the No/Cancel button event.

I kinda feel like that "OK" should rather trigger the "Confirmed" event.
So for Yes/No you got both events and for "OK" only the "Confirmed".

skittles1:
Changed the OK button to use the Confirmed event and adjusted the new login error dialogs to match (amended the commit).

PR #221, opened 2017-04-01
Add stat change UI.

skittles1:
Added a stat change UI, incremented Ogre client min version. StatChangeInfo now keeps track of the original levels sent by the server, and uses those as a maximum for the UI.


skittles1:
Hey, have you had a chance to look over this yet?

cyberjunk:
I haven't built it yet, but I reviewed it and it seems fine the way you implemented it.
I can merge now, if it's working fine for you?

cyberjunk:
Please also increment the corelib dll version.
There are some changes to StateChangeInfo class there.

skittles1:
Incremented corelib dll version. Probably safe to merge, though nobody else apart from me has tested it out. I know @MorbusM59 wanted to adjust some of the UI elements before we release it so that would be easier once merged.

PR #240, opened 2017-09-18
Convert old mail timestamps to unix time.

skittles1:
Convert any old mail timestamps (on load) from offset time to the proper unix epoch time and save the mail after conversion.

Change the OFFSET value to 0 as unix time with no offset will be used for mail/news.

TODO: wondering where in the 'mail received' process I should set IsTimestampUpdated to true (with the assumption that any mail received has a unix timestamp). One way to verify the timestamp is valid, if we didn't want to assume it was, is to check against MAX_KOD_INT as any timestamp from the server larger than that would be a valid updated one (or from the early 1970s but I don't think we need to account for that). We could get around needing an IsTimestampUpdated field by performing the same check on mails when fixing (e.g. lower than MAX_KOD_INT? add offset and save).

### DIFF
--- a/Meridian59/Data/Models/Mail.cs
+++ b/Meridian59/Data/Models/Mail.cs
@@ -47,11 +47,13 @@ namespace Meridian59.Data.Models
         public const string TAG_MAIL        = "mail";
         public const string TAG_RECIPIENTS  = "recipients";
         public const string TAG_RECIPIENT   = "recipient";
-        public const string TAG_TEXT        = "text";       
+        public const string TAG_TEXT        = "text";
         public const string ATTRIB_SENDER   = "sender";
         public const string ATTRIB_TIMESTAMP= "timestamp";
         public const string ATTRIB_NAME     = "name";
         public const string ATTRIB_TITLE    = "title";
+        public const string ATTRIB_ISTIMESTAMPUPDATED = "istimestampupdated";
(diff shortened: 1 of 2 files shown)


Meridian59/Data/Models/Mail.cs
```diff
@@ -337,6 +359,7 @@ public void ReadXml(XmlReader Reader)
             // read attributes
             Sender = Reader[ATTRIB_SENDER];
             Timestamp = Convert.ToUInt32(Reader[ATTRIB_TIMESTAMP]);
+            IsTimestampUpdated = Convert.ToBoolean(Reader[ATTRIB_ISTIMESTAMPUPDATED]);
```
cyberjunk: Are you sure this works when the xml atribute "istimestampupdated" is missing, like in old mails?
I kinda expected that you would have to use boolean.TryParse() and a default of false

Meridian59/Data/Models/Mail.cs
```diff
@@ -181,6 +183,7 @@ public byte[] Bytes
         protected readonly List<string> recipients = new List<string>();
         protected ServerString message;
         protected string title;
+        protected bool istimestampupdated;
```
cyberjunk: Please rename this to "isTimestampUpdated"

PR #244, opened 2017-11-20
Add guild "invite" chat command.

skittles1:
Allows users to send a guild invite by using the chat window to "invite" a player to their guild.

### DIFF
--- a/Meridian59/Client/BaseClient.cs
+++ b/Meridian59/Client/BaseClient.cs
@@ -1103,6 +1103,19 @@ namespace Meridian59.Client
         }
 
         /// <summary>
+        /// Tries to invite the specified player (by object ID) into your guild
+        /// </summary>
+        public virtual void SendUserCommandGuildInvite(uint Member)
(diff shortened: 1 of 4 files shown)


Meridian59/Client/BaseClient.cs
```diff
@@ -1102,6 +1102,21 @@ public virtual void SendUserCommandGuildInvite()
             }
         }
 
+#if !VANILLA && !OPENMERIDIAN
+        /// <summary>
+        /// Like SendUserCommandGuildInvite, but called via "invite" chat command.
+        /// </summary>
+        public virtual void SendUserCommandGuildInviteChat(uint Member)
```
cyberjunk: Please remove the if vanilla.. checking around this function and rename it to "SendUserCommandGuildInvite()" (yes, same name). In fact this is just an alternative variant which allows providing a ID directly rather than using the current target. There are several other cases where equally named function allow to do something with different params.

PR #247, opened 2017-11-22
Add CharInfoNotOk improvements, Confirm dialog box adjustment for more text.

skittles1:
Handle the new CharInfoNotOk error codes returned by the server to better describe char creation errors.

Changed the error size to byte from the previous code I had up.

Not sure if there is a way to skip the switch statement in UIAvatarCreateWizard.cpp where it converts from the CharInfoNotOkError enum to the language enum, or if this way is preferred for safety.

Also the OK dialog box's Text field isn't big enough to hold these error messages, I'm still looking for a good way to adjust this box (and possibly move the OK button depending on size) without making the defaults huge or adding a new box to the .layout file. This topic http://cegui.org.uk/forum/viewtopic.php?f=5&t=7101 implies there might not be an easy CEGUI method to do so.

### DIFF
--- a/Meridian59/Protocol/GameMessages/GameMode/CharInfoNotOkMessage.cs
+++ b/Meridian59/Protocol/GameMessages/GameMode/CharInfoNotOkMessage.cs
@@ -16,14 +16,66 @@
 
 using System;
 using Meridian59.Protocol.Enums;
+using Meridian59.Common.Constants;
+using Meridian59.Common.Enums;
 
 namespace Meridian59.Protocol.GameMessages
 {
+    /// <summary>
+    /// Tells the client about errors during character creation.
+    /// </summary>
     [Serializable]
     public class CharInfoNotOkMessage : GameModeMessage
-    {        
-        public CharInfoNotOkMessage()
-            : base(MessageTypeGameMode.CharInfoNotOk) { }
+    {
+        #region IByteSerializable implementation
+        public override int ByteLength
+        {
+            get
+            {
+#if VANILLA || OPENMERIDIAN
+                return base.ByteLength;
+#else
+                return base.ByteLength + TypeSizes.BYTE;
+#endif
+            }
+        }
+
+        public override int WriteTo(byte[] Buffer, int StartIndex = 0)
+        {
+            int cursor = StartIndex;
+
+            cursor += base.WriteTo(Buffer, StartIndex);
+#if !VANILLA && !OPENMERIDIAN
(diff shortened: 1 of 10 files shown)


cyberjunk:
Looks quite good so far...

1) The switch statement seems fine to me. The alternative would be using the corelib enum value directly as index into EN_CHARINFONOTOKERROR_OKDIALOG[] respectively the DE version arrays (= as language string ID). But I wouldn't do that...

2) CEGUI indeed doesn't support autosizing to make containers show all contents. But I think it should be possible to increase the popup size in general, if it's too small. I can remember looking into measuring text widths/heights in order to write some manual adjustements of container sizes, but I think it was quite a pain. Another option is to reduce the text-size. Also possible like -1 fontsize per 100 characters/words etc.

3) I think VANILLA support can probably added to this by working inside the game-message class only.
Rather than reading the byte it should probably just set the value to the generic error or the previously used charactername in use error.

skittles1:
Added some if guards - is this what you meant?

Meridian59/Protocol/GameMessages/GameMode/CharInfoNotOkMessage.cs
```diff
@@ -16,14 +16,62 @@
...
-    {        
-        public CharInfoNotOkMessage()
-            : base(MessageTypeGameMode.CharInfoNotOk) { }
+    {
+        #region IByteSerializable implementation
+        public override int ByteLength
+        {
+            get
+            {
+                return base.ByteLength + TypeSizes.BYTE;
```
cyberjunk: You need the if vanilla checking here too (no additional byte?)

cyberjunk:
>> Added some if guards - is this what you meant?

Yes, that's what I had in mind for backward compatibility. Just see comments at ByteLength return

skittles1:
I think with the dialog box maybe the best option is to have two versions of the OK box (assuming this one will have the most varied text): version 1 is the existing single-line OK box, version 2 is 3 or 4 lines. I think the code could decide which box to spawn based on the length of the string to display.

skittles1:
So another option, I've managed to get the box formatting so that it can display between 1-6 lines of text.

6 lines:


1 line:


Think this looks okay?

cyberjunk:
>> Think this looks okay?

Yes, it does. In fact not totally pleasant with that much of text, but it's ok.

Anything else you want to add here?

It's still marked as WIP

skittles1:
Nothing else to add, rebased and WIP removed.

PR #249, opened 2017-11-28
Add RoomContentsFlags message to update ObjectFlags for objects in room, ChangeObjectFlags message to update ObjectFlags for a single object in room or inventory.

skittles1:
#### Commit 1
New protocol message to handle cases where the ObjectFlags of objects in a room change, but overlays etc do not, so resetting all data for the object is unnecessary. Server can send this for the full list of objects in the room, or a partial list.

#### Commit 2
Added a protocol message to handle a single object's ObjectFlags, basically like ChangeMessage but just sending object ID and ObjectFlags.

#### TODO:
ObjectUpdate (BP_CHANGE) has some special handling for ObjectFlags.DrawingType.SecondTrans: https://github.com/cyberjunk/meridian59-dotnet/blob/33209cf1c141de217b048d4498fc5559791c4b03/Meridian59/Data/Models/ObjectUpdate.cs#L142-L145
The new message could also change DrawingType to SecondTrans, so this probably needs to be handled somewhere, possibly in UpdateFromModel.


skittles1:
Looking more into the SecondTrans drawingtype, I think it's going to be tricky to handle it in Ogre without introducing a property like the classic client has (object_node's secondtranslation is always set in classic server.c to FILTERWHITE90), or a property to save the initially sent ColorTranslation for an object which can be set back if it receives updated flags that remove the SecondTrans effect.

I guess an alternative would be to change the getter for ColorTranslation to return ColorTransformation.FILTERWHITE90 instead of ColorTranslation if the object/projectile/overlay in question has DrawingType.SecondTrans. If we did this, we could also remove the parts in ObjectUpdate etc where ColorTranslation is set.

PR #255, opened 2017-12-05
IRCBot changes.

skittles1:
Added/adjusted functionality in IRCBot.

#### Commit 1
- Added a new config field, AllowedUsers, which can broadcast but not use bot admin commands. Removed the +O IRC flag requirement, users just need to be registered. The 'WhoX' (who extended) IRC protocol handles this by returning the registration status of users as they join, and the bot keeps a list of them.
- Added a RecentAdmins list to keep track of admins who have recently used a (server) admin command. Only send the 'garbage collect' message to those that have (reduce spam). Reset this every server save.
- Added DisplayMessages property, can toggle whether game content is displayed to IRC (e.g. if we need to turn it off for a bit to stop spamming the channel).
- Moved some code to a Util class.

#### Commit 2
Add an updated IrcDotNet lib with added WhoX protocol, built with the code from [this branch](https://github.com/skittles1/IrcDotNet/tree/whox_rollback) (I added this to the library). I don't know if we want the .xml file, I can remove it if not.

#### Commit 3
Added some admin commands to the bot, available via PM. Can list users/admins/registered users, perform registration checks, toggle game output to IRC and add/remove users. Hope I'm on the right track with the class structure (trying to loosely mirror ChatCommand with a bit less abstraction).

#### Commit 4
Added a manager class for WhoX queries. Will only send one query per second, and waits 5 seconds after receiving the query before sending it. Prevents flooding the server with WhoX queries, and gives users time to register nick after joining the channel.

#### TODO (done now)
- Delay the WhoX registration check for a bit on login, since some clients don't register straight away (and some users do it manually). AFAIK there isn't a message sent out when a user registers if they do not receive OPs. (done, commit 4)
- Remove the Thread.Sleep() I'm using to delay the WhoX checks when the bot logs in. Some form of delay is required else the bot will spam reg checks on all users and get throttled by Esper.NET. Time checks could probably take care of both these issues. (done, commit 4)


skittles1:
No more features to add on this PR, any issues with the changes?

cyberjunk:
I think the ircdotnet version which I used is included in the repo here in "Tools/ircdotnet".
I haven't monitored development on the lib in the last years, but back then there were serious multithreading issues in the original code...

skittles1:
I'm not sure if those issues are fixed, the lib versioning is still 0.x.x so issues might be expected. I'm not actually using their most recent code, they changed the lib to target .NET standard 1.5 and that doesn't seem to play well with our .NET framework project (without adding a ton of extra DLLs). Should I include the branch I'm building with (with rolled back .NET standard changes) in "Tools/ircdotnet"?

cyberjunk:
On first glance the, the old version in "Tools/ircdotnet" and the one you linked seem to be quite different. It's possible they fixed the issues from back then, but I didn't check completely..
You're basically free to use whatever version you want.

skittles1:
Added the IrcDotNet source (from https://github.com/skittles1/IrcDotNet/tree/whox_rollback)

PR #256, opened 2017-12-11
Remove daylight savings from GetMeridianTime().

skittles1:
The servers will be using UTC hours and minutes to set game time, so
that every even (UTC) hour is midnight. This is more easily reproducible
by the clients and means daylight savings doesn't need to be handled.


cyberjunk:
Was this changed at some point? I think I remember adding this because the timer got off by one hour.

PR #257, opened 2017-12-24
Split large ServerStrings before sending to IRC channel.

skittles1:
EsperNET has a ~450 char limit which is lowered by using color/style codes, and messages from the server (i.e. broadcasts) can often be in excess of this (up to 550 IIRC). Split at a conservative point (280 chars) to allow for color codes and send the message to the channel in multiple parts. Also attempt to split at a logical point (whitespace) near 280 chars, to avoid chopping up the IRC color/style codes.

Increment assembly version, update configuration.xml for testing server/channel.

Note: I added the try/catch block with exception logging because we're currently using this as-is, some testing has been done but haven't figured out if there's any rare edge cases not being handled.

### DIFF
--- a/Meridian59.Bot.IRC/IRCBotClient.cs
+++ b/Meridian59.Bot.IRC/IRCBotClient.cs
@@ -178,6 +178,51 @@ namespace Meridian59.Bot.IRC
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


cyberjunk:
Cool. I must have missed this one.
Is it ready for merge?

skittles1:
Should be okay to merge.

PR #260, opened 2018-02-13
Add more textures for grass decoration.

skittles1:
Added some extra textures to grass.xml so they receive grass decoration.


cyberjunk:
Nice. Thx.

PR #264, opened 2018-06-11
UI visibility changes.

skittles1:
Add missing UI visibility checks which allow hiding of avatar, action
grid, left/right buttons and room enchantments. Visibility for these
items can be set in configuration.xml.

Fixed incorrect XML element being checked for action grid visibility.


cyberjunk:
I think these windows can't be hidden by default.
Wonder why they even have a config parameter for their visibility...

PR #266, opened 2018-08-15
Allow .NET patcher to update classic client.

skittles1:
- Generalised some of the patcher features to allow patching classic client in addition to Ogre.
- Simplfied the infobox string addition code - calling function now provides the whole string to display.
- Patcher now checks file hashes against local files in a separate thread (FileScanner class) before beginning download. This allows the correct size of the total download to be displayed in the UI, and no 'already downloaded' files are shown as downloading.

#### TODO
I've left the file hash checking in the Worker class also as a kind of fallback (basically catch edge cases like users moving files in during the download) but it is mostly superfluous and we could probably remove it.
Still want to test a little more, we've used it to update classic a couple times but still need to update Ogre using it and also try a fresh install.

### DIFF
--- a/Meridian59.Patcher/DownloadForm.cs
+++ b/Meridian59.Patcher/DownloadForm.cs
@@ -7,20 +7,19 @@ namespace Meridian59.Patcher
 {
     public partial class DownloadForm : Form
     {
-        private readonly List<PatchFile> files;
         private readonly LanguageHandler languageHandler;
+        private readonly PatchDownloadStats patchDownloadStats;
         private string appendLog;
         private double lastTick;
-        private long lastLengthDone;
 
         /// <summary>
         /// Constructor
         /// </summary>
         /// <param name="files"></param>
-        public DownloadForm(List<PatchFile> files, LanguageHandler languageHandler)
+        public DownloadForm(LanguageHandler languageHandler, PatchDownloadStats patchDownloadStats)
         {
             this.languageHandler = languageHandler;
-            this.files = files;
+            this.patchDownloadStats = patchDownloadStats;
             InitializeComponent();
             CenterToScreen();
         }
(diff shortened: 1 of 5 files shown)


cyberjunk:
>> Patcher now checks file hashes against local files in a separate thread (FileScanner class) before beginning download. This allows the correct size of the total download to be displayed in the UI, and no 'already downloaded' files are shown as downloading.

To be honest, I don't like this part:

There were 8 threads (interchangable by replacing a single value) before caluclating MD5 hashes in parallel. Now there is a bottleneck with only a single thread doing it...

What is the actual benefit of this?

Also if you carefully look at the values in the patcher, none should say anything about "Download", like "Download Speed", they just say "Speed", which in fact can be considered like "Processing Speed", "Patching Speed" or similar.

Going to test this out on the weekend, but I might request you to revert this back to the original fully multithreaded solution.

cyberjunk:
PS: If you absolutely want to calculate all the MD5 hashes before the download, then do it like this:

1) Add a second threadsafe queue
2) Remove the download (second step) when threads run over the first/existing queue
3) Instead of the download, let them add the processed items which must be downloaded into the new, second queue
4) Once the first queue is empty (all MD5 calculated), make all the threads process the second queue (download)

something like this...

skittles1:
> What is the actual benefit of this?
Also if you carefully look at the values in the patcher, none should say anything about "Download", like "Download Speed", they just say "Speed", which in fact can be considered like "Processing Speed", "Patching Speed" or similar.

We thought it would be more useful for players if the update showed them only the number of files they needed to update and the size of just those files as the goal, rather than using the whole-client values there. That meant hashing as a first step and I wanted to get it off the UI thread to avoid locking the UI during the hashing process.

I much prefer your solution of using the existing Workers and two queues, I can rewrite it that way.

cyberjunk:
>> We thought it would be more useful for players if the update showed them only the number of files they needed to update and the size of just those files as the goal, rather than using the whole-client values there. 

Sure, why not. But I wouldn't really consider this such a major improvement and it raises a lot questions. How to deal with the progress bar then? Will there be two progress bars after each other? First one for hashing, then one for the downloading? Or still a cumulated one using the overall byte sizes?

>> ..and I wanted to get it off the UI thread to avoid locking the UI during the hashing process.

Yes, definitely shouldn't be done by UI thread.

>> I much prefer your solution of using the existing Workers and two queues, I can rewrite it that way.

Yes, please. If you need some assistance let me know. You can also adjust the PR and keep this topic out of scope for now. Just focus on getting the original client compatibility for now and leave the threading/patching as it is.

skittles1:
I rewrote it using the worker class to scan the files prior to downloading.

I also moved all the stats to a new class for (theoretically) easy access between the threads but in practice even with the locks the values seem to get messed up, e.g. incrementing PatchDownloadStats.TotalFilesChecked vs calling IncrementFiles() with 8 threads running will cause TotalFilesChecked to be 5-15 files short when accessing the property directly. My solution was going to be to add functions to perform whatever accesses necessary but perhaps i'm just doing something wrong with the lock objects? Property access works correctly with 2 worker threads but breaks on 4 and 8 threads.

cyberjunk:
AFAIK "Interlocked" is not a real lock. It's "just" atomic...
I also think it's not perfect to have a single value/lock and then so many threads bailing for it.

I added the lock to the PatchFile (which makes it in fact many many locks) for a reason:
What you actually need to synchronize is only the mainthread and one worker at a time, only that worker the mainthread is currently "wanting"/reading something from (in this case the value of 'LengthDone').

I think you might want to add a property like "LengthHashed" to the PatchFile.cs class and make it work like the "LengthDone" (which you might want to rename into "LengthDownloaded" or keep as is). Make sure to use a second lockobject (not the existing).

The worker modify "LengthHashed" when they hashed the file, they modify "LengthDone" (or LengthDownloaded) when they download the file. And the UI thread gathers together the stats from all the workers like before.

cyberjunk:
>> I think you might want to add a property like "LengthHashed" to the PatchFile.cs 

Oh, and additionally I think you need another new property like "LengthOld". Because the current 'Length' of PatchFile is the new file size. To make use of the 'LengthHashed' regarding retrieving an overall progress/done percentage for hashing, you would use like: sum(LengthHashed)/sum(LengthOld), where the overall progress for download would be still: sum(LengthDownloaded)/sum(LengthNew), or with current names rather: sum(LengthDone)/sum(Length).

skittles1:
Outdated, closing this one.

PR #267, opened 2018-08-18
Allow .NET patcher to update classic client (split from earlier PR).

skittles1:
Added classic client patching to the dotnet patcher. Patcher differentiates between '6-arg classic' and '7-arg classic' - for a short time the classic client passed 6 arguments to club before a 7th was added, very few clients would really be updating this way but better to handle that edge case.

Other misc changes:
- Simplfied the infobox string addition code - calling function now provides the whole string to display.
- Added messagebox after patching finishes stating whether anything was updated. Form needs to be updated prior to this in case it misses the last downloaded bytes.
- Incremented patcher version.

### DIFF
--- a/Meridian59.Patcher/Patcher.cs
+++ b/Meridian59.Patcher/Patcher.cs
@@ -8,6 +8,7 @@ using System.Net;
 using System.Reflection;
 using System.Runtime.Serialization.Json;
 using System.Security.Principal;
+using System.Text.RegularExpressions;
 using System.Threading;
 using System.Windows.Forms;
 
@@ -19,9 +20,27 @@ namespace Meridian59.Patcher
     /// </summary>
     public static class Patcher
     {
+        private enum UpdateFormat
+        {
+            None,
+            DotNet,
+            Classic6Arg,
+            Classic7Arg,
+            ClassicManual
+        };
+
         private const int NUMWORKERS        = 8;
         private const int MAXRETRIES        = 3;
-        private const string URLDATAFILE    = "patchurl.txt";
+
+        private const string CLASSICURLDATAFILE = "dlinfo.txt";
+        private const string CLASSICEXENAME     = "meridian.exe";
+        private const string CLASSICPROCESS     = "meridian";
+        private const string CLASSICPATCHEREXENAME = "club.exe";
+
+        // Fallback default for old 6-arg classic client, expected to be very low prevalence.
+        private const string CLASSICDEFAULTJSON = "clientpatch.txt";
+
+        private const string DOTNETURLDATAFILE = "patchurl.txt";
         private const string CLIENTEXENAME  = "Meridian59.Ogre.Client.exe";
         private const string CLIENTPROCESS  = "Meridian59.Ogre.Client";
         private const string PATCHEREXENAME = "Meridian59.Patcher.exe";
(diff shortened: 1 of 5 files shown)


Meridian59.Patcher/Patcher.cs
```diff
@@ -237,6 +282,14 @@ private static void ProcessQueues()
         /// <param name="args"></param>
         private static void ReadCommandLineArguments(string[] args)
         {
+            // Handle classic client arguments.
+            if ((args.Length == 6 || args.Length == 7)
+                && args[1].Contains("UPDATE"))
+            {
+                ReadCommandLineArgumentsClassic(args);
```
cyberjunk: Please don't call 'Classic' variants from inside 'DotNet'/'Ogre' variants of functions.
Decide which one you want before the call is made...

Meridian59.Patcher/Patcher.cs
```diff
@@ -258,34 +311,91 @@ private static void ReadCommandLineArguments(string[] args)
...
+        private static void ReadCommandLineArgumentsClassic(string[] args)
+        {
+            // Classic client has two possible argument formats - both are a single string
+            // with data in quotes except for an unquoted UPDATE in the second position.
+            // The early classic updater contained 5 quoted fields, the modern one contains 6.
+            // The difference is the last two arguments, 6 arg is missing patch file name
+            //(e.g. clientpatch.txt) which is the second-last argument in 7 arg.
+
+            // Probably need to kill the update process here.
+            if (args.Length < 6 || args.Length > 7)
```
cyberjunk: Basically the same here: Decide which version of the function is the right one.
Then assume the correct function to be called inside them.
Remove this check, it's guaranteed to have 6 or 7 args length if it's called and the return handling doesn't make it much better in case it's incorrectly used.

Meridian59.Patcher/Patcher.cs
```diff
@@ -19,9 +20,26 @@ namespace Meridian59.Patcher
...
+        };
+
         private const int NUMWORKERS        = 8;
         private const int MAXRETRIES        = 3;
-        private const string URLDATAFILE    = "patchurl.txt";
+
+        private const string CLASSICURLDATAFILE = "dlinfo.txt";
+        private const string CLASSICEXENAME     = "meridian.exe";
+        private const string CLASSICPROCESS     = "meridian";
+        private const string CLASSICPATCHEREXENAME = "club.exe";
```
cyberjunk: May be want to use this constant below in 'EXCLUSIONS' ?
Because it still says 'club.exe' there.

Meridian59.Patcher/Patcher.cs
```diff
@@ -60,6 +79,10 @@ public static class Patcher
         private static string baseUrl    = "";
         private static string jsonUrl    = "";
         
+        private static UpdateFormat updateFormat = UpdateFormat.None;
```
cyberjunk: If I get this right we currently decide which mode the patcher runs solely based on the amount of command line paramters that have been passed. Means passing 6 or 7 args will assume it's classic, while any other amount will make it think it's ogre/dotnet. IMO that's a very poor solution which can cause issues later (e.g. assume for some reason, two modes need the same amount of arguments etc.).

I think there are only 2 clean solutions for this:

1) Use preprocessor macros to build different binary versions for classic/ogre. The overhead for building it twice should be acceptable.

2) Pass a commandline argument when starting it which expliclity tells which mode it should run in

Meridian59.Patcher/Patcher.cs
```diff
@@ -70,7 +93,15 @@ static void Main(string[] args)
             watch.Start();
 
             // parse commandline arguments
-            ReadCommandLineArguments(args);
+            if ((args.Length == 6 || args.Length == 7)
```
cyberjunk: Please use this if branching also to set "updateFormat" already.
Then inside your "ReadCommandLineArgumentsClassic" use the enumeration types for v6/v7 to decide which way to go.

cyberjunk:
👍

PR #268, opened 2018-08-19
Show scanning/hashing and downloading differently in updater UI.

skittles1:
Draft implementation of providing UI feedback on hashing process. Shows num_hashed/total_files and length_hashed/total_length in UI prior to download starting. I added a Thread.Sleep() in Worker in the hashing section so that we can see the hashing part of the UI in testing (otherwise it won't show up at all on my computer).

Added a bool to track the start of downloading in Patcher so that 'starting download' text can be displayed in the form (not sure if there's a better way to track that) and also prevented files from giving the 'downloaded file.ext' text in form if they weren't actually downloaded.

Removed extra form tick after loop - issue with progressbar text was due to requiring Text to be modified prior to the Value assignment.

### DIFF
--- a/Meridian59.Patcher/Patcher.cs
+++ b/Meridian59.Patcher/Patcher.cs
@@ -73,10 +73,11 @@ namespace Meridian59.Patcher
         private static readonly Stopwatch watch            = new Stopwatch();
         private static readonly WebClient webClient        = new WebClient();
 
+        // Track progress on JSON file download in a threadsafe/clean way.
+        private static readonly JsonFileProgress jsonFileProgress = new JsonFileProgress();
+
         private static DownloadForm form = null;
         private static int filesDone     = 0;
-        private static bool abort        = false;
-        private static bool isRunning    = true;
         private static bool isHeadless   = false;
         private static string baseUrl    = "";
         private static string jsonUrl    = "";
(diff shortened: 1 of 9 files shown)


skittles1:
I added a draft of the UpdateStage handling in the 2nd commit, still have a few more things I want to do and test (and still needs German strings for the status changes).

Meridian59.Patcher/Patcher.cs
```diff
@@ -286,6 +296,60 @@ static void Main(string[] args)
...
+
+                    x64Process = new Process();
+                    x64Process.StartInfo = pi;
+                    x64Process.Start();
+                }
+                else
+                    x64NgenDone = true;
+            }
+            else if (x64Process.HasExited == true)
+                x64NgenDone = true;
```
cyberjunk: I think you could try to additionally examine the "ExitCode" here to check if NGEN was successful.

Meridian59.Patcher/Patcher.cs
```diff
@@ -165,21 +187,43 @@ static void Main(string[] args)
             // start download of patchinfo.txt
 
             if (!isHeadless)
-                form.DisplayStatus(1, languageHandler.DownloadingPatch);
+                form.DisplayStatus(String.Format(languageHandler.DownloadingPatch, updateFormat == UpdateFormat.DotNet ? 4 : 3));
```
cyberjunk: This expression:

updateFormat == UpdateFormat.DotNet ? 4 : 3

is used quite often.
I think it would be cleaner to add a bool property like "IsDotNet" to Patcher with a getter that includes this expression.
So if we ever get more than 4 or 3 steps we don'T have to adjust so often.

Meridian59.Patcher/Patcher.cs
skittles1: I'm not sure if the user can do anything if it's unsuccessful, but should we report the 'finished' text as "Update complete"/"Update complete, but optimization unsuccessful"?

cyberjunk: Good question. I think my original idea was like a warning popup telling the user, that startup optimization was not successful. But that does not help much anyways...

PR #269, opened 2018-08-20
Fix irregular right-harnd border on patcher UI.

skittles1:
Using FormBorderStyles FixedSingle or Fixed3D causes the right-hand
border of the UI to stick out a few pixels past the title bar. Using
Sizeable instead (the default I guess since it didn't add a property for
it in designer) fixes this issue. Added a minimum size so the UI can't
actually be resized.

The other changes were made by VS when i changed the border, seems
that's the way it wants it laid out.

### DIFF
--- a/Meridian59.Patcher/DownloadForm.Designer.cs
+++ b/Meridian59.Patcher/DownloadForm.Designer.cs
@@ -29,21 +29,9 @@
         private void InitializeComponent()
         {
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(DownloadForm));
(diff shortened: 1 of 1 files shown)


cyberjunk:
>> The other changes were made by VS when i changed the border, seems
that's the way it wants it laid out.

You can modify this Designer file also manually. You gotta be careful of course to not break it (not really suited for adding new controls etc.), but if it's only for changing a single value or so I usually do it that way.

cyberjunk:
Hmm. Did the old variant also show the "resize" mouse cursor?
Cause this variant does not allow you to resize, but shows the mouse cursor suggesting to be able to do so.. :-(

PR #271, opened 2018-08-22
Switch updater from MD5 to SHA256, hashing/finishing transition pause, translations.

skittles1:
Added a pause in-between hashing and downloading in a way that would keep ticking the UI to prevent unresponsiveness, set to 40 ticks by default.

Earlier I mentioned the issue Judere was having with FIPS (see [here](http://blog.aggregatedintelligence.com/2007/10/fips-validated-cryptographic-algorithms.html) for info). Two solutions are possible for this:
   1. Catch the exception and tell the user to disable FIPS (and hopefully convince them it is safe to do so)
   2. Switch to a hashing algo which won't trigger this exception

I prefer the second solution since it's way easier for users and means less support needed from us. I've included both solutions in this PR for now, also useful if you wanted to test the FIPS issue locally.

EDIT:
Switched it to SHA256 but left in an exception handler for rare cases where Win XP might be outdated/bugged and not be able to create the SHA256 hasher. Rare, but better we catch it than crash.

Removed the 'client updated' messagebox and instead put that info in the status label and keep the form open ~1.5sec.

Compressed the splash image for the picturebox, halves exe size.


cyberjunk:
>> Earlier I mentioned the issue Judere was having with FIPS (see here for info). 

Interesting, I wasn't aware of this FIPS thing so far.
Is it a Windows XP only thing or does it also affect Windows 7 and above?

Because OgreClient uses the MD5 class to hash the password for login and if I understand you correctly this could cause a crash for some players on login?

skittles1:
Judere specifically turned it on, from what I've read it will almost always be off though we've had a few players with unexplained crashes on startup (can't recall any recently). This was also on his XP machine so he hasn't tried Ogre there. I think we should probably switch the password hashing too, but that's a bit more involved than switching the file hashing algorithm.

Meridian59.Patcher/Patcher.cs
```diff
@@ -209,6 +216,7 @@ static void Main(string[] args)
             ///////////////////////////////////////////////////////////////////////
             // mainthread loop
 
+            ushort transitionTicks = 0;
```
cyberjunk: I think I don't really like this 'transitionTicks' concept. See, if someone got a slow system and is already spending like 3 seconds in the hashing step, he'll spend another 1s (or whatever it is right now) during the forced wait.

As suggested earlier, I think you should be saving a tick from the watch, when you enter e.g. the HasingFiles state. Then compare the current ticks against this saved tick while you're in the 'HashingTransition' state to figure out, if it's at least N miliseconds ago. This will fix the issue with the delay being applied on cases that already took much longer anyways.

Meridian59.Patcher/Patcher.cs
```diff
@@ -653,18 +666,32 @@ private static void OnWebClientDownloadDataCompleted(object sender, DownloadData
...
-                    workers[i].Start();
+                // Catch potential exception if the SHA256 hasher couldn't be created.
+                // Some installs of Win XP may have a bug in their .NET framework that
+                // will cause the hasher not to be found, unlikely we will hit this case
+                // but rather catch it than crash. On our single test XP SP3 system, we
+                // did not hit this exception.
+                try
+                {
+                    // create worker-instances and start them
+                    for (int i = 0; i < workers.Length; i++)
```
cyberjunk: I suggest to move the worker instantiations to an early position in the main[]. This way you don't need to download the the patchinfo etc. until you figure out, that the system does not support the hashing in the try{}

You could leave the ".Start()" here, but actually I think you can also move this too, because it doesn't hurt if the workers start earlierer (e.g. before entering the mainthread loop). I think (and they'are also stopped from the main[]) and this callback here doesn't really seem a nice place to do all this....

PR #286, opened 2019-01-26
Add NPC quest UI.

skittles1:
Add all the quest system upgrades.

Easiest to read this by looking at each commit separately, with commit message detailing contents.

I also added a gitignore entry for the NuGet packages folder as I have a bunch of stuff in there locally.


cyberjunk:
@skittles1 
Don't forget to also increment version in OgreClient/app.rc

skittles1:
Updated app.rc in the version commit.

PR #297, opened 2019-06-02
Framework for active skills, spell + skill addition/removal protocol changes.

skittles1:
#### Commit 1
- Add corelib code for active skills (mirroring how spells are handled in more places).
- Spell and skill removal now remove the ability from the UI lists and SpellObjects/SkillObjects lists, also fixes a bug where spells were not removed from the UI on spell removal server-side (skills worked because the whole list got resent).
- BP_STAT/Stat protocol can now handle addition of a single spell or skill, so the server can send just that one and not the whole list when the player gets an ability added.

#### Commit 2
- Add a ReqPerform protocol mirroring ReqCast but for (active) skills. Can be triggered using a chat command.

#### Commit 3
- Skills UI now allows dragging (only) active skills onto the action grid so they can be performed by hotkey. Active skills use the spell icon type to accomplish this, with passive skills using the old skill icon.
- Skills UI now behaves more like Spells UI (catching input, selectable items, double-click to perform skill, right-click to view description).

### DIFF
--- a/Meridian59.Ogre.Client/UISkills.cpp
+++ b/Meridian59.Ogre.Client/UISkills.cpp
@@ -20,7 +20,7 @@ namespace Meridian59 { namespace Ogre
       Window->subscribeEvent(CEGUI::FrameWindow::EventCloseClicked, CEGUI::Event::Subscriber(UICallbacks::OnWindowClosed));
 
       // subscribe keyup
-      Window->subscribeEvent(CEGUI::FrameWindow::EventKeyUp, CEGUI::Event::Subscriber(UICallbacks::OnKeyUp));
+      Window->subscribeEvent(CEGUI::FrameWindow::EventKeyUp, CEGUI::Event::Subscriber(UICallbacks::Skills::OnKeyUp));
    };
 
    void ControllerUI::Skills::Destroy()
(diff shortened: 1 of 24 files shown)


cyberjunk:
I think quite a lot of what is affected by this might be already implemented in the C++ port.
I'll have to carefully go over it and implement the changes there too, at least on classes I considered done already. Hope I'll be able to finish that at latest next weekend (will try tomorrow).

cyberjunk:
I've ported some of this to the C++ port, but not yet 100% finished.
I'm also struggling a bit with the whole thing.
Just to make that clear: I totally like the idea of active weapon skills.

I just feel like we're inreasing the semantic overlapping of Spells, Skills and Actions and overcomplicating things may be...

If I compare that to games like WoW, they don't differ between spell abilites and skill abilities. They simply have active and passive abilities, some are spell-casts, some are physical abilities (depending on your  class/build)

I don't have much idea of our server-side regarding this, but I feel like at some point we should think about merging spells and skills and just think of them as "abilities" (either passive or active) and which would all work the same for the client (e.g. regarding execution in console, protocol and actionbuttons etc.), instead of pretty much duplicating everything.

Meridian59.Ogre.Client/UISkills.cpp
```diff
@@ -58,9 +58,32 @@ namespace Meridian59 { namespace Ogre
       CEGUI::WindowManager& wndMgr = CEGUI::WindowManager::getSingleton();
       StatList^ obj = OgreClient::Singleton->Data->AvatarSkills[Index];
 
+      SkillObject^ skillObj =
+         OgreClient::Singleton->Data->SkillObjects->GetItemByName(obj->ResourceName, false);
```
cyberjunk: Any reason you do a lookup by name here instead of object id?
Stat list has object id attached and so does skill object.

PR #299, opened 2019-06-12
Allow Invalidate to close specific ConfirmPopup windows.

### LINKED ISSUE #296: ConfirmPopup windows stay open through a system save.
Windows with data (e.g. stat change UI) get closed during a save when the data is cleared, however the popups (e.g. stat change confirmation) stay open and in the case of stat change, can be used to send invalid data to the server. I'm not sure if there's a simple way to notify the client from DataController's Invalidate() to remove any ConfirmPopup instances since there's not any coupling there, but I'll be looking into this ASAP.

skittles1:
ConfirmPopup windows don't get closed on data invalidation (i.e. system
save) which can cause an issue with stat reset - invalid stat reset data
could be sent to the server even though the stat reset window itself has
been hidden.

This change clears the data in ConfirmPopup and hides it if present when
Invalidate is called and any open ConfirmPopup has the CloseOnInvalidate
bool set to true. Currently set for stat reset, guild, quest and char
create popups.

~~This change clears the data in ConfirmPopup and hides it if present when~~
~~the visible status on the stat reset window changes from visible to~~
~~hidden.~~

~~I think a more elegant way to handle this would be to inform the~~
~~ConfirmPopup window directly from DataController's Invalidate method,~~
~~maybe by having ConfirmPopup handle events on a new property in~~
~~DataController (something like a DataInvalidated boolean that gets~~
~~toggled at the start/end of Invalidate()). Let me know if you'd like~~
~~something like that instead.~~

Fixes #296

### DIFF
--- a/Meridian59/Data/DataController.cs
+++ b/Meridian59/Data/DataController.cs
@@ -995,7 +995,7 @@ namespace Meridian59.Data
         /// Invalidates all data not valid anymore
         /// after server-save cycle
         /// </summary>
-        public void Invalidate()
+        protected virtual void Invalidate()
         {
             // clear lists
             RoomObjects.Clear();
--- a/Meridian59.Ogre.Client/ControllerUI.h
+++ b/Meridian59.Ogre.Client/ControllerUI.h
@@ -1234,16 +1234,18 @@ namespace Meridian59 { namespace Ogre
          static ::CEGUI::PushButton* LargeOK = nullptr;
          static uint ID = 0;
          static DialogMode Mode;
+         static bool CloseOnInvalidate;
 
          static void Initialize();
          static void Destroy();
          static void ApplyLanguage();
-         static void ShowChoice(const ::CEGUI::String& text, uint id);
-         static void ShowOK(const ::CEGUI::String& text, uint id);
-         static void ShowChoiceLarge(const ::CEGUI::String& text, uint id);
-         static void ShowOKLarge(const ::CEGUI::String& text, uint id);
+         static void ShowChoice(const ::CEGUI::String& text, uint id, bool closeOnInvalidate);
+         static void ShowOK(const ::CEGUI::String& text, uint id, bool closeOnInvalidate);
+         static void ShowChoiceLarge(const ::CEGUI::String& text, uint id, bool closeOnInvalidate);
+         static void ShowOKLarge(const ::CEGUI::String& text, uint id, bool closeOnInvalidate);
          static void _RaiseConfirm();
          static void _RaiseCancel();
+         static void DataInvalidated();
       };
 
       /// <summary>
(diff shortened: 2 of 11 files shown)


cyberjunk:
Reviewed this a bit.

What I don't really l ike is that it seems any open ConfirmPopup always gets closed as long as StatChangeWizward was "visible". Since StatChange wizard is not modal, I think a player can open the stat change wizard, and then cause a totally different popup, that should not get closed on serversaves/if statchange gets closed.

Regarding:

>> I think a more elegant way to handle this would be to inform the
ConfirmPopup window directly from DataController's Invalidate method,
maybe by having ConfirmPopup handle events on a new property in
DataController (something like a DataInvalidated boolean that gets
toggled at the start/end of Invalidate()). Let me know if you'd like
something like that instead.

Possible, but there's another way to do this. The DataController in the core library is just the base class, the one actually used in OgreClient is "DataControllerOgre" in the OgreClient project. This already overrides a few methods, so you could simply make "Invalidate()" (or whichever you want) virtual in the base class and then override it with some additional code. From there you can also access the static UI such as the ConfirmPopup. But an event is also fine...

I think we might want to store a new boolean value with the confirmpopup, like "CloseOnInvalidate". This needs to be set by the ConfirmPopup user, like the others properties that get set. And a new method like "Invalidate()" that checks "CloseOnInvalidate" and if true, does things like your ClearData().  That Invalidate() would either be called from the event you mention or from the overriden Invalidate() of DataControllerOgre.

skittles1:
Thanks, not sure how I missed the DataControllerOgre class :) Will redo using the override.

skittles1:
Updated the PR description and commit with the Invalidate override method.

Meridian59/Data/DataController.cs
```diff
@@ -995,7 +995,7 @@ public void Reset()
         /// Invalidates all data not valid anymore
         /// after server-save cycle
         /// </summary>
-        public void Invalidate()
+        protected virtual void Invalidate()
```
cyberjunk: Why did you change this from public to protected?

cyberjunk:
Nice, that's exactly what I had in mind.
Just curious why you changed the Invalidate() from public to protected, was that by purpose?

cyberjunk:
Also, please check ControllerUI::IsInitialized in DataControllerOgre Invalidate() before you call DataInvalidated() on the ConfirmPopup UI window. Likely not a problem right now, but the DataController gets alive a lot earlier than the UI (and might also live during an engine/UI/window reload etc. when the UI is invalid)

cyberjunk:
Going to merge and fix the minor things myself.

cyberjunk:
See: https://github.com/cyberjunk/meridian59-dotnet/commit/94d40a7d7b3b20c350987b8cfc173054c6e51616

PR #300, opened 2019-06-13
Allow left double-click to activate targeted object.

### LINKED ISSUE #17: Double-Clicking Mana Nodes / Levers doesnt activate them
I dont know if thats really an issue, but it could be a problem for vets who are using the new client and have to figure out stuff like that. Maybe it's possible to make a double-click feature like in the vanilla client? 

Things like Brax Levers, Mana Nodes, Raza/Marion Levers...

skittles1:
Catch time of last left click (mouseup) and if another left click comes
within 350ms, register as a left double click and try activate the
current highlighted object (previous click would have selected it).
Allows using mana nodes/levers and looking inside chests with double
click.

Closes #17

### DIFF
--- a/Meridian59.Ogre.Client/ControllerInput.cpp
+++ b/Meridian59.Ogre.Client/ControllerInput.cpp
@@ -12,6 +12,7 @@ namespace Meridian59 { namespace Ogre
 
       tickMouseDownLeft    = 0;
       tickMouseDownRight   = 0;
+      tickMouseClickedLeft = 0;
 
       isCameraFirstPerson  = true;
       isInitialized        = false;
@@ -104,6 +105,7 @@ namespace Meridian59 { namespace Ogre
 
       tickMouseDownLeft  = 0;
       tickMouseDownRight = 0;
+      tickMouseClickedLeft = 0;
 
       isCameraFirstPerson  = true;
       isInitialized        = false;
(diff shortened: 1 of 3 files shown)


cyberjunk:
In general, I like this idea. It's also possible, since the double-click action (activate) includes the single-click action (select). This fact is important because the single-click is always performed first or else has to be delayed (see bad inventory select click delay)

What I don't like is the huge copy+paste of the already quite ugly "PerformMouseOver" :-)
PerformMouseOver already got a boolean "isClick". I think I would prefer if you switch that from bool to int and rename it to "clicks".  Then 0 is the mouse-over, 1 the first click (single) and 2 the second click within interval (double). It will make the function even a bit more ugly but I guess still better than the huge copy+paste.

Also the clicks are not really guaranteed to be counted on the same object. While it's theoretically possible the first click was on A and the second was on B, I think this is quite unlikely and probably does not need special code to deal with.

skittles1:
I'll change it to use the existing PerformMouseOver.

> While it's theoretically possible the first click was on A and the second was on B, I think this is quite unlikely and probably does not need special code to deal with.

In the current changeset this is dealt with by checking the 2nd click has landed on the current target (which should have been set by the 1st click) so clicking on a different target on either click should prevent an activation being sent.

skittles1:
Done, I also reduced the nesting a bit in PerformMouseOver making it hopefully a bit less ugly.

It does still handle the case of clicking quickly on 2 different objects (not counted as activation click) but it's possible to select a target, then double-click on (1st click) empty space then (2nd click) targeted object and activate it. I think that's probably expected behavior, but if not I could change it by keeping track of the object clicked first.

cyberjunk:
Nice. Thanks.

PR #302, opened 2019-06-18
Resize MultiLineEditboxes to fit text where possible.

### LINKED ISSUE #31: UI Problem: Description Window isnt resizing to the size of its content
I have to resize the window everytime i click on an item to be able to read the whole text. ogre should do that on its own if there's enough space on the view port to show everything.

image

skittles1:
Object descriptions (mainly for items) and spell descriptions tend to
force a scrollbar on the inspection window as the window isn't large
enough to display all the text. At the same time, a default larger
window would look strange for short descriptions. This change
dynamically resizes the window so more of the text will fit, from the
current min up to the current max size.

Object inspection windows with inscriptions aren't resized as this is
less of a problem, and the main focus (inscription editbox) on the most
common items with inscriptions already uses the available max space.

Closes #31.

One question about this - instead of calling SetLayout() for a changed
Description in UIObjectDetails.cpp, it could instead check if the object
has an inscription and call Window->setHeight without all the other
adjustments. Think that is better than the way it is now?

### DIFF
--- a/Meridian59.Ogre.Client/Constants.h
+++ b/Meridian59.Ogre.Client/Constants.h
@@ -129,6 +129,7 @@
 #define UI_FILE_LAYOUT                 "Meridian59.layout"
 #define UI_FILE_ANIMATIONS             "animations.xml"
 #define UI_DEFAULTPADDING              5
+#define UI_DESCRIPTIONPADDING          7
 #define UI_MINIMAP_TEXNAME             "CEGUI/MiniMap"
 #define UI_MOUSE_SINGLECLICKTIMEOUT    1.0f
 #define UI_AVATARNAME_FOR_UNSET        "<< EMPTY >>"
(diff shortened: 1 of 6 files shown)


cyberjunk:
One thing:

I think I would prefer if your new 'GetAdjustedWindowHeightWithMLEB()' would be inside ControllerUI. There are already some small CEGUI related helper functions there, like 'IsRecursiveChildOf()' or 'SetVUMeterColorFromProgress()'. And since it's exclusive to CEGUI (params, output and code uses nothing else and is called from nowhere else), I feel this would be the better place.

>> One question about this - instead of calling SetLayout() for a changed
Description in UIObjectDetails.cpp, it could instead check if the object
has an inscription and call Window->setHeight without all the other
adjustments. Think that is better than the way it is now?

Not sure if I understand it, if it creates same behaviour with better code I'm always a fan of it :-)

skittles1:
Moved the function to ControllerUI and fixed up the SetLayout() call.

cyberjunk:
Thanks

PR #304, opened 2019-08-01
Allow shift-tab from login menu, increment ogre vers.

skittles1:
- Pressing shift-tab in the password field in the login menu will now move back to the username field.
- Increment Ogre client version.


cyberjunk:
Please also update the version in OgreClient app.rc file (it's stored twice)
And increment the corelib version...

skittles1:
Updated corelib vers, as far as I can tell all references in app.rc are updated.

PR #307, opened 2019-08-09
Sort char creation ability lists, raise points to 125.

skittles1:
- Char creation ability points raised to 125.
- Spells and skills now sorted in char creation, with school listed. Sorting is by school, then by level, then by ability name.
- Display ability selection errors as a popup in the char creation UI, letting the user know what went wrong.

#### Logic changes to ability selection:
- Disallow selecting both Qor and Shal'ille spells together.
- Require taking two level 1 abilities to take a level 2 ability in the same school.

Also as an aside, I considered combining as much as I could into 'Ability' stuff rather than Spell/Skill duplicates, but there's some differences in what gets sent for AvatarCreatorSpellObject and AvatarCreatorSkillObject, and if we combine it then it'd be tricky to have it work on other servers (possible but clunky).


skittles1:
This should be ready to review/merge now.

cyberjunk:
I think I would prefer if you do this without LINQ.  I don't remember exactly but there was a reason that I did not use it so far and I would prefer if it stays that way.

skittles1:
Replaced the LINQ calls with loops.

PR #308, opened 2019-08-25
Inventory swap bug fix.

skittles1:
When swapping inventory items, need to start from count - 2 as the swap is with the next element in the list.

I incremented the ogre client minor version, should I also increment the 1.0.5.x version? It's at 1.0.5.9, wondering if you wanted 1.0.6.0 or 1.0.5.10 there.


cyberjunk:
>> should I also increment the 1.0.5.x version

Yes, please. I don't like two binary different files to have the same assembly version.

PR #313, opened 2020-04-10
Allow users to slightly increase client brightness.

skittles1:
Adds an slider to the Options UI under the Engine tab allowing the user
to alter the ambient + directional brightness. The lowest setting is the
current brightness, and max is 1.8x the current brightness. On the
highest setting, Ogre displays rooms more closely to how classic does
(though classic tends to be too bright).

This is in response to some players complaining that Ogre graphics are
much too dark. Not everyone feels this way, so a slider allowing the players
to pick is a compromise.


Meridian59.Ogre.Client/OgreClientConfig.h
```diff
@@ -51,6 +51,7 @@ namespace Meridian59 { namespace Ogre
       literal bool         DEFAULTVAL_ENGINE_DISABLE3DMODELS         = false;
       literal bool         DEFAULTVAL_ENGINE_DISABLENEWSKY           = false;
       literal bool         DEFAULTVAL_ENGINE_DISABLEWEATHEREFFECTS   = false;
+      literal float        DEFAULTVAL_ENGINE_BRIGHTNESSFACTOR        = 0.0f;
```
cyberjunk, line 54: Default should be 1.0 (light as before), not 0.0 (no ambient) ?

skittles1: The UI slider has to have a minimum of 0.0 so I have the factor between 0.0-0.8 and add 1.0 to it when it is used.

cyberjunk: Ok. I see, I missed +1.0f in:

>> float brightnessFactor = Math::Clamp(OgreClient::Singleton->Config->BrightnessFactor + 1.0f, 1.0f, 1.8f);

I think it's a bit misleading. It's rather a 'BrightnessFactorOffset' (added to the Factor) this way.
Any chance you can make BrightnessFactor  (=the slider) in between 1.0 and 1.8 and just skip the +1.0?

cyberjunk:
Also one other thing:

It seems you apply this only on the ambient light, not the directional light?
Have you tried using like smaller value, but on both?

I just fear a bit that it will make ambient light so much brighter that there will be no more recognizable directional light any more...

If you want to keep it for ambient light only, then please change the name to "AmbientLightFactor".

skittles1:
The directional lighting is very dim normally:

That's the view with ambient turned off and looking directly where the sun is illuminating. The better option might be to apply the factor to the directional lighting also.

Meridian59.Ogre.Client/OgreClientConfig.h
skittles1: Sorry missed this comment, the CEGUI sliders must have a minimum value of 0 - the docstring above CEGUI::Slider::setMaxValue says `set the maximum value for the slider.  Note that the minimum value is fixed at 0.`. Found that when I couldn't find a setMinValue as I wanted to have it set up that way (1.0-1.8).

PR #321, opened 2020-06-20
Add new chat colors, refactor chat color handling, sound rolloff adjustment.

skittles1:
- Add new chat text colors.
- Assign the chat color char values in the enum rather than storing them
separately (assigns the int value, important for comparisons).
- Move GetChatString from both UIChat and UINPCQuestList to Util to
avoid duplication.
- Lower the Ogre sound rolloff value as players were complaining that
sounds were being attenuated too quickly with distance.
- Increment core lib/ogre versions.


skittles1:
Just bumping this PR, we've had these changes live on the 105 & 112 versions of Ogre client since about August so it should be okay to merge unless you notice any style issues.

PR #328, opened 2022-01-30
better phase visuals

andygeiss:
I have added a new UI notification to display "PHASED" instead of "PARALYZED", because as players mentioned before they have noticed practically no difference (little phase icon is maybe not obvious enough) in the heat of the battle.
Thus players thought they are already phased but instead they were only held and died ;-)


skittles1:
On testing this one, I found the buffs were being iterated very frequently because `OnDataPropertyChanged` is called a lot (basically any time any property in DataController is modified), thus `ControllerUI::SplashNotifier::OnDataPropertyChanged` was not an ideal place to be handling the phase notification. In #337 I added a listener to the `AvatarBuffs` list which calls a new function which checks if the single element added or removed is the phase buff. This PR can be closed now.

PR #329, opened 2022-02-04
action bar added

andygeiss:
Ive added a 5th action bar to increase quality of life for mages (4 and 5 schoolers)
to make things easier.


Meridian59/Data/Lists/ActionButtonList.cs
```diff
@@ -155,6 +155,19 @@ public static ActionButtonList DEFAULT
                 buttons.Add(new ActionButtonConfig(46, ActionButtonType.Spell, Common.Constants.ResourceStrings.Spells.Riija.PHASE));
                 buttons.Add(new ActionButtonConfig(47, ActionButtonType.Action, AvatarAction.Inspect.ToString().ToLower()));
 
+                buttons.Add(new ActionButtonConfig(48, ActionButtonType.Unset, String.Empty));
+                buttons.Add(new ActionButtonConfig(49, ActionButtonType.Unset, String.Empty));
+                buttons.Add(new ActionButtonConfig(51, ActionButtonType.Unset, String.Empty));
```
skittles1, line 160: 50 is missing here, and you have 60 as the last one instead of 59 (this one is 0-based).

skittles1:
This change works with a fresh client which creates its own configuration.xml (with the above ActionButtonConfig fix) but doesn't work with existing installations which have config files with 48 action buttons.

skittles1:
I've fixed up the action button PR and submitted it in #334, so this can be closed now.

PR #332, opened 2022-04-19
WIP: Experimenting with targeting adjustments.

### LINKED ISSUE #322: Ogre Camera & Targeting Priorities.
Would generally like to close the gap of classic/ogre skill basis removing the incredible demand for "Entity" lists in a players toolbelt and going back to the origins of targeting through brackets (or tab on ogre)

Targeting in its current state is roughly 340 degrees as compared to a normal MMO/FPS style game where your perspective is like an isosceles triangle, the tip being the player extending outwards approx 45-60 degrees. Currently it feels as if, and even proven to be able to target someone outside your FoV that would be considered behind you. Attached image shows a more realistic approach to player targeting revisions.

Also a secondary approach would be even to add a "Targeting Priority" system where we can just control the flow of what we're targeting. IE: xx Priority: Nearest, Priority: Furthest, Priority: Enemy, Priority: Ally. Available in a drop down menu in the settings, if thats a possible solution.

ogretargetings

skittles1:
Attempting to fix #322.

The issue is that when attacking without a target selected, the client prefers closer objects even if they are behind the player or outside a reasonable field of view. I've tested setting a lower FoV but it's not perfect yet even at 135 degrees (down from 180) and there's still some edge cases where an undesired object will be picked on the edge of FoV.

Ideally we could draw the player's FoV into the graphics window both for debugging and so players could turn it on if they want to see what they can actually attack, whether that be a shader on the ground or an Ogre light set to spotlight centered on the player & directed forwards (though this isn't exactly frustum-shaped unless the spotlight is behind the player a bit). What do you think would be the best method here? Still looking into the specifics of both of these but thought I'd open the PR early in case you have any advice.


skittles1:
Current changes are live but users still reporting that attacking without a target will try to attack objects behind them, so re-added WIP to the title.

PR #339, opened 2022-12-07
dont close loot-list on get/get all

m59solo:
Just an obvious suggestion


m59solo:
back to github school xD

